Issue #33, opened 2019-05-02
Create Application Area Workflow


chiatt:
This should include the following workflow steps:

- [ ] Application Area Address (move from the consultations workflow)
- [ ] Application Area Name (move from the consultations workflow)
- [ ] Area Map (Just stub this step out with a new-tile-step until the mapping workflow step is completed)
- [ ] Area Description
- [ ] Area Designation

Consult the latest mockups and @dwuthrich to confirm the content of each step.

chiatt:
Use the workflow stubbed out in #28 if #28 is already completed

chiatt:
Meets requirements

PR #42, opened 2019-05-15
creates new workflow for Application Area, includes some steps moved …

whatisgalen:
…out of Consultation workflow, re #33


chiatt (approved):
Look good!

PR #366, opened 2019-10-17
365 django2.2 update

apeters:


apeters:
don't want to merge into master

PR #433, opened 2020-01-28
First Push of pc-pkg-changes

Philbox:
All nodes checked for correct vocab. All vocabularies created


Philbox:
Ignore the settings.py file.

Issue #460, opened 2020-03-02
es index default mapping in arches-her/datatypes/bngcentrepoint prints "DOING GREAT STUFF"


whatisgalen:
one of the custom datatypes (bngcentrepoint) in the project prints the statement in its `default_es_mapping` method. It should be removed or changed to something more informative

aj-he:
PR submitted

PR #461, opened 2020-03-04
460 bngdatatype esmap

aj-he:
Updated the default-es-mapping #460 

Done in both the function and the pkg.


aj-he:
@whatisgalen as requested!

PR #470, opened 2020-04-01
Updates from Phil

aj-he:


whatisgalen:
A few comments here:

- I'm seeing a photo widget (`js`, `htm`, `json`) in both `arches-her/widgets` and `pkg/extensions/widgets` that may actually be redundant with the photo-gallery card component which has recently been moved to core arches. Can you check whether the core arches photo-gallery component fulfills the need? 

- in `pkg/business_data` there are a 2 sub-directories containing resource instance data and associated files: `successfully loaded` and `to load`. Files placed in these subdirectories will not be found by the `import_business_data` command thus resource instances in csv's contained therein will not be loaded. Can you confirm this is the desired behavior? Furthermore, some files which had previously successfully loaded resource instances (ActorOrg, ActorPerson, Consultation, Application Area) have been moved into either of those two subdirectories and thus no longer load their resource instance data.

- in general, `.csv` files which do not have a paired `.mapping` file (e.g. `example1.csv` and `example1.mapping`) cannot be loaded. It appears there are quite a few in the `to load` directory. Please do reach out if you need further information on this.

- Some resource instance csv files cannot be loaded due to a missing (or misplaced) graph/resource json file. They are: `Activity_Core.csv` and `Activity_Core2.csv`. Note that if these files share the same graph/resource model, they should be combined into a single `.csv` file with a single paired `.mapping` file. (This would also apply to what looks to be a single type of resource instance data split up among 16 `Event_CoreXX.csv` files in the `to load` directory.)

whatisgalen (changes requested):
please see comment on url name change in `pkg/extensions/search/search-results-consultations/search-results-consultations.js` and general comments on the PR itself

Philbox:
> A few comments here:
> 
> * I'm seeing a photo widget (`js`, `htm`, `json`) in both `arches-her/widgets` and `pkg/extensions/widgets` that may actually be redundant with the photo-gallery card component which has recently been moved to core arches. Can you check whether the core arches photo-gallery component fulfills the need?

It probably would. I haven't added these so will check where they came from

> * in `pkg/business_data` there are a 2 sub-directories containing resource instance data and associated files: `successfully loaded` and `to load`. Files placed in these subdirectories will not be found by the `import_business_data` command thus resource instances in csv's contained therein will not be loaded. Can you confirm this is the desired behavior? Furthermore, some files which had previously successfully loaded resource instances (ActorOrg, ActorPerson, Consultation, Application Area) have been moved into either of those two subdirectories and thus no longer load their resource instance data.

This is the desired behaviour. As I'm working on the models I want the instance to be empty and I'm using these folders to store files that I know do/will load. I will remove the folders and only add the business data to be loaded into the main business data folder.

> * in general, `.csv` files which do not have a paired `.mapping` file (e.g. `example1.csv` and `example1.mapping`) cannot be loaded. It appears there are quite a few in the `to load` directory. Please do reach out if you need further information on this.

I am aware of this. I will create the relevant .mapping files when I load the data.

> * Some resource instance csv files cannot be loaded due to a missing (or misplaced) graph/resource json file. They are: `Activity_Core.csv` and `Activity_Core2.csv`. Note that if these files share the same graph/resource model, they should be combined into a single `.csv` file with a single paired `.mapping` file. (This would also apply to what looks to be a single type of resource instance data split up among 16 `Event_CoreXX.csv` files in the `to load` directory.)

These files have been broken down as various attempts to load instance data with more than 1 value for a node (ie. a Primary Name and an Alternative Name) has failed and will only load the first row. Having checked the cardinality of the cards I don't understand why it won't load so have separated the data and am appending. I'd be happy to be directed as to how to get the data to load without breaking it.

aj-he:
@Philbox I can see you now have a branch in the main arches-her repo. Are you happy for me to close this PR then?

PR #526, opened 2020-09-29
Prevents error in the map popup within the map card. re archesproject/arches#6659

chiatt:
Moves permission check/error handling logic to the view model to prevent error in the template. re archesproject/arches#6659


chrabyrd (approved):
lgtm

PR #542, opened 2021-02-12
541 update bng files to v5.1

phudson-he:


chiatt (approved):
@phudson-he, sorry I missed this notification and didn't get this taken care of sooner. This looks good. You might consider writing some of those large arrays and dictionaries into a module so that they don't interrupt the logic so much.

PR #554, opened 2021-02-26
Remove the gap between consultation dates in workflow, #552

### LINKED ISSUE #552: Reduce space between consultation dates

njkim:
Update the css to hide an element and remove the gap, #552 
Resolve an issue with consultations dates step (unnecessary gap between dates)


chiatt (approved):
Works as expected

Issue #556, opened 2021-03-09
Replace 'Completion Date' with 'Target Date' in Consultation Dates step


chiatt:
Replace 'Completion Date' with 'Target Date' in the Consultation Dates step of the Consultation workflow.

chiatt:
Resolved in #557

PR #557, opened 2021-03-10
Show target date rather than completion date in workflow

chiatt:
Hide completion date rather than target date re #556


njkim (approved):
Works as expected

PR #561, opened 2021-03-18
Fix issue with active consultations not zooming to consulations geom,…

### LINKED ISSUE #555: Active consultations maps do not zoom to geometry
Each map on the page should automatically zoom to it's corresponding consultation geometry.

ryan86:
… re #555


chiatt (approved):
Resolves #555

PR #564, opened 2021-03-19
Upgrade correspondence workflow,  #538 & #442

### LINKED ISSUE #442: Update docx templating with new models

### LINKED ISSUE #538: Update the Correspondence workflow to support the latest graphs

njkim:
Upgrade the correspondence workflow, #538 
Also update the letter templates #442


chiatt (approved):
Works as expected. We may need to hide the template tags when the value is null, but I've written a separate ticket for that. #565

Issue #536, opened 2021-02-03
Update the Communication workflow to support the latest graphs 


chiatt:

njkim:
The nodegroup/node ids are updated. (pr #540 )

There are some remaining issues including,
The big one card should be separated to small workflow steps.

njkim:
The communication workflow may be good one to implement "component based workflow"

PR #567, opened 2021-03-19
Communication workflow updated for new graphs, #536

njkim:
Communication workflow is updated to use the digital objects related-resource, #536 
Also one card is split into multiple steps by selectively hiding widgets.


chiatt (approved):
resolves 536

PR #568, opened 2021-03-22
Migrate ref numbers step to component-based-step

chiatt:
Migrates ref numbers step to component-based-step


njkim (approved):
works as expected

Issue #559, opened 2021-03-16
Add instruction panel to workflow steps


chiatt:
The instruction panel for tabbed workflows is missing from most steps.
- [x] Consultation
- [x] Application Area
- [x] Site Visit
- [x] Communication
- [x] Correspondence

PR #570, opened 2021-03-22
559 information panels

chiatt:
Add info panels to workflows


chiatt (commented):
Resolves #559

njkim (commented):
Works as expected

Issue #578, opened 2021-03-25
Update workflow summary step - Application Area


dwuthrich:
The Arches-HER workflows should be updated so that the final step (eg: summary step) matches the mockups attached[https://app.zenhub.com/files/180664470/5306fbf8-a592-4e97-84c8-93e0148201f1/download](https://app.zenhub.com/files/180664470/5306fbf8-a592-4e97-84c8-93e0148201f1/download)

chiatt:
Complete for all workflows

PR #590, opened 2021-04-07
Move shared logic into summary-step and add related resources

chiatt:
Move shared logic into summary-step and add related resources to summary. re #578


njkim (approved):
Works as expected

PR #593, opened 2021-04-10
Update wastebin, arches#7282

njkim:
Update wastebin to reflect what in the current step's wastebin properly in the custom steps, archesproject/arches#7282


chiatt (approved):
Ensures wastebin is properly updated

Issue #587, opened 2021-04-06
Update the summary step for the Communication workflow


njkim:
Update the final step of the Communication workflow so that the summary wii show as in the mockup

PR #594, opened 2021-04-12
Make sure to use the consultation resourceid, #587

njkim:
Add the restore the resource instance id of the consultation resource to make sure display the right summary, #587


chiatt (approved):
Resolves #587

Issue #600, opened 2021-04-14
The upload-document step in communication workflow does not load properly when revisited


njkim:
After a user uploaded the document(s), save and move forwards and then comes back to the step, the page does not load at all and does not show the previously uploaded document.

PR #601, opened 2021-04-14
Update upload-step to load properly on revisit, #600

njkim:
Update the upload-step in communication workflow to load the page and previously uploaded document, #600


chiatt (approved):
Resolves #600

Issue #598, opened 2021-04-12
Communication workflow final step creates an unrelated resource instance


chrabyrd:
Depends on (https://github.com/archesproject/arches/pull/7308 + https://github.com/archesproject/arches-her/pull/597)

The consultation workflow final step creates a second resource instance without the data that's been built over the course of the workflow. The data is saved in an original resource instance.

As the final-step is being refactored anyhow, this is a good opportunity to make sure we have proper functionality.

To reproduce:
1. Delete all `Consultation` resource instances
2. Navigate through the `Consulation` workflow
3. (Easiest way if via filtered search) Notice how two resource instances are created.

The workflow should create a single valid resource containing all data captured within the workflow

chiatt:
I'm no longer able to reproduce this issue.

PR #602, opened 2021-04-14
Finishes consultation final step

chiatt:
Finishes consultation final step and refactors app area final step to handle for cases when tiles are missing. re #598


njkim (approved):
Works as expected

PR #612, opened 2021-04-19
update summary step returned array, #607

### LINKED ISSUE #607: The workflows are updated based on the updated label based graph endpoint
The summary steps of the workflows should be updated based on the updated label based graph endpoint which returns an array for the cardinality n card even when there is only one value.

njkim:
Update summary step logic based on the new resources api, #607 
The new api returns the array for the cardinality of "n" even there is only one value.


chiatt (approved):
Fixes #607

PR #614, opened 2021-04-20
Update the consultation status, #611

### LINKED ISSUE #611: Update consultation status in the consultation workflow
When a user creates a consultation in consultation workflow, the consultation status needs to be set to active in final step of the workflow.

njkim:
Update the consultation status to active at the end of the workflow, #611
It checks the status to avoid creating multiple tiles.


chiatt (approved):
Fixes #611

Issue #588, opened 2021-04-06
Update the summary step for the Correspondence workflow 


njkim:
Update the final step of the Correspondence workflow so that the summary wii show as in the mockup

PR #616, opened 2021-04-20
Add a resource editor link in the summary step, #605

### LINKED ISSUE #605: In final steps, add link to navigate back to resource editor
Users should be able to navigate back to the resource editor to make additional change to the specified resource instance.

njkim:
Add a resource editor link of the resource being edited and/or created in the workflow summary page, #605


chiatt (approved):
Resolves #605 although we will still need to implement this in the correspondence final step as part of #588

PR #618, opened 2021-04-21
Fix null in address string, #617

### LINKED ISSUE #617: App-area name which is composed from address sting should not include null

njkim:
Compose address string without null as a string, #617


chiatt (approved):
Resolves #617

Issue #613, opened 2021-04-19
site-visit workflow does not always create a data in a single parent tile.


njkim:
If you already have more than one site-visit tile, the new child tiles are often not created under the new parent tile.
This will result in the recommendation, observation, photos, attendees which are attached to the wrong site-visit.

chiatt:
Resolved in #619

PR #619, opened 2021-04-22
Update site-visit workflow to track the parent tileid, #613

njkim:
Update the site-vist workflow to track the parent tile id , #613

Linked to the update of the component based step, archesproject/arches#7340


chiatt (approved):
Resolves #613

Issue #620, opened 2021-04-27
Update BNG functions and datatype to work with stable/5.2.x


csmith-he:
Some minor changes are required to the BNG datatype and the BNG to GeoJSON function in order to work in a stable/5.2.x environment.

HE to supply the update

dwuthrich:
completed

PR #621, opened 2021-04-27
BNG function and datatype updates for 5.2.x (#620)

csmith-he:
Updating the BNG datatype and BNG to GeoJSON function to work with Arches stable/5.2.x as per issue #620 

Added strict=false to validate in the BNG datatype
Changed SQL in BNG to GeoJSON function to SELECT * FROM refresh_geojson_geometries();
Manual testing with these changes applied - all resource models in arches-her that use the BNG datatype checked

### DIFF
--- a/arches_her/datatypes/bngcentrepoint.py
+++ b/arches_her/datatypes/bngcentrepoint.py
@@ -19,7 +19,7 @@ details = {
 
 
 class BNGCentreDataType(BaseDataType):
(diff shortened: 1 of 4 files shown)


csmith-he:
@chiatt I've been asked to flag this up for your attention.  Just a couple of minor updates for the BNG stuff to work in 5.2.x

chiatt (approved):
This changes look good.

PR #622, opened 2021-04-27
Update correspondence workflow final step, re #588

### LINKED ISSUE #588: Update the summary step for the Correspondence workflow

ryan86:


chiatt (approved):
Resolves #588

PR #623, opened 2021-04-28
Add query string to redirect, arches#7322

njkim:
Fix the redirecting middleware to add query string when the web page is redirecting to /arches-her/ address
Fix the issue of related resource button is not working, archesproject/arches#7322


chiatt (approved):
Fixes issue with report not loading in related resource info panel. re #7322

Issue #581, opened 2021-03-29
Consultations Workflow: application areas overlay map pop-up does not show "select" link


dwuthrich:
In the consultations workflow, the first step requires a user to select an application area to relate to the consultation. One way the user should be able to select the appropriate app area is by clicking on it on the map.  To do so, the map pop-up should include a "select" link in the footer that will add the application area as a related resource to the consultation

chiatt:
@dwuthrich This should be fixed in the latest, but let me know if you don't see the link.

PR #626, opened 2021-04-30
Update related resource min zoom level, #581

njkim:
Update the related resource map layer minimum zoom level in consultation workflow to make sure to show the app-area layer, #581


chiatt (approved):
Resolves #581

PR #636, opened 2021-05-14
Update external data access between steps

chiatt:
Updates external data access between steps to support changes in this PR: #7422


njkim (approved):
Works as expected

Issue #606, opened 2021-04-16
Application area workflow name step creating multiple names in one workflow


njkim:
If you finish and save the address step with "Make the Area Name the same as the Area Address" checked, and come back and edit the address, it will create a new tile for the name instead of modifying the existing name.

PR #646, opened 2021-05-28
Update app area address-name steps, #606

njkim:
Update the application area address & name steps, #606
Combine two steps in one.


chiatt (approved):
Works as expected, but 'Undo' needs to be implemented: https://github.com/archesproject/arches-her/issues/650

PR #649, opened 2021-05-28
Update custom search to correspond to changes in core arches, 

### LINKED ISSUE #648: `options.loading` undefined error when loading search page
Error when loading the search page prevents the search results from appearing in the search result list.

chiatt:
re archesproject/arches#7439


njkim (approved):
Works as expected

Issue #651, opened 2021-05-29
Some basemaps cause related resource map card to fail


chiatt:
The Emerald, Dark, Light, and Satellite Streets layers all cause the related resource map card to fail when a user tries to perform a buffer or intersection. They need to be replaced with modern versions that do not cause the same errors.

chiatt:
Resolved in #652

PR #652, opened 2021-05-29
Update basemaps with related resource map compatible versions, re #651

chiatt:
Updates Dark, Light, and Satellite Streets and removes Emerald (replaced by new streets layer in core Arches).


njkim (approved):
All the base maps work great.

Issue #645, opened 2021-05-28
Related condition widget fails in first step of the communication workflow


chiatt:
![image.png](https://images.zenhubusercontent.com/57d0a2a4e40e5714b16e73d7/fb015cff-bff6-4ab8-95fd-01f9c0aa3b35)

chiatt:
Resolved in #653

PR #653, opened 2021-06-02
Fix node value dropdown in first communication workflow step

chiatt:
re #645

Must be QA'd with https://github.com/archesproject/arches/pull/7481


njkim (approved):
works as expected

PR #664, opened 2021-06-23
Add Resource Instance Name in final summary step, #628

### LINKED ISSUE #628: Workflow summary: Show name of instances created in addition to model type
The summary page provides a link to the instance(s) created in the workflow, but only shows the model type.  We should also include the instance name.  In the example below, the "Created or Modified Resource Instance" list should show instance using the following pattern "Instance Name (model type)"

njkim:
Add Resource Instance Name in final summary step, #628


arches_her/media/js/views/components/workflows/upload-document-step.js
```diff
@@ -115,6 +109,23 @@ define([
             params.resourceid(ko.unwrap(this.consultationResourceId));
             params.workflow.resourceId(ko.unwrap(this.consultationResourceId));
     
+            params.defineStateProperties = function(){
```
chiatt: defineStateProperties should probably just be removed here

chiatt (changes requested):

arches_her/media/js/views/components/workflows/upload-document-step.js
njkim: The branch is updated the most recent master as of 12/10 and some minor fixe was added.

njkim: This is ready to be reviewed again.

chiatt (approved):
Looks good to me.

PR #665, opened 2021-07-02
remove extraneous yarn.lock file

mradamcox:
Removes a yarn.lock file that was likely created by running `yarn install` in the wrong directory.


chiatt (approved):
Looks good, thanks @mradamcox

PR #668, opened 2021-07-13
Adds docker files/compose for arches-her

### LINKED ISSUE #667: Create docker compose setup for development
It would be nice to have a docker compose setup (similar to what we have in afs) for arches-her.

aarongundel:
* Adds docker files/compose for arches-her
* Also some auto-formatting for the arches_her settings.py file

closes #667


chiatt (approved):
works as expected

PR #677, opened 2021-07-29
666 pc graph rdm updates

Philbox:


aj-he:
@dwuthrich FYI, I've loaded this package into our dev environment and there are no reported errors.

dwuthrich:
@Philbox Could you please address the following issues:
> The Artifact model uses a different location branch that all the other models.  Can we model the "Discovery Method" and "finder" card as top-level cards instead?
> Why does the HLC model use a different pattern for locations?  Is it important enough to break the pattern used for most other models?
> Metatype nodes are all visible in the scientific dates card.  Please update the branch and all models that use the branch
> The "Standard Deviation Type" node is a collection that points to a single value of... Standard deviation.  Should this field be a numeric value instead?
> Make sure that the external cross reference and system reference number cards in the same order in each model
> Heritage Area can only have 1 parent area.  Same with Parent Organizations in the Organization model.  Is this correct? (Heritage Assets can have multiple parents)
> Upgrade configuration of designation/protection card's designation mapping node to default to London (or at least England) instead of the whole world
> Why does the Period model use a different pattern for names (e.g.: a separate card for alternate names)?  Is this important enough to break the pattern used by all the other models?
> Why does the Place model use a different version of the location branch?  Is it important enough to be different than the other models?

Philbox:
@dwuthrich Please review latest modifications

njkim (approved):
Error occurs in url datatype but appears unrelated to graph changes

PR #698, opened 2021-08-26
Add koop config to pkg and to docker setup and remove cantaloupe

ryan86:


chiatt (approved):
Adds koop as expected

PR #702, opened 2021-09-03
fixes docker postgres docker volumes not storing data

aarongundel:
fixes pathing so docker volumes store postgres data.  Temp fix until shared volumes are in use.


chiatt (approved):
Looks good to me

chiatt (approved):
Looks good to me

Issue #707, opened 2021-09-13
Migrate consultation workflow to new workflow architecture


chiatt:

chiatt:
Fixed in #723

PR #723, opened 2021-11-19
Migrate consultations workflow to v6 workflows, re #707

chiatt:
Updates the consultations workflow to be compatible with v6 workflows


njkim (approved):
The workflow is working fine with most recent arches workflow logic.
The may be some minor remaining issues which may require some core arches update in the component abstract such as multi tile step.

PR #724, opened 2021-11-20
Migrate application area workflow to v6 workflows, re #708

### LINKED ISSUE #708: Migrate application area to new workflow architecture

chiatt:
Migrates application area workflow to v6 workflows, re #708


njkim (approved):
The workflow is working great with the new workflow logic.

Issue #710, opened 2021-09-13
Migrate the correspondence workflow to new workflow architecture


chiatt:

chiatt:
Resolved in #726

PR #726, opened 2021-11-24
Update correspondence workflow, #710

njkim:
The workflow is updated to work with the new workflow logic, #710


chiatt (approved):
Works as expected

PR #727, opened 2021-12-02
Report template for the consultation resource

njkim:


arches_her/media/js/views/components/reports/consultation.js
```diff
@@ -14,31 +14,250 @@ define([
...
+            };
+
+            const communicationsNode = self.getRawNodeValue(self.resource(), 'communications');
+            if(Array.isArray(communicationsNode)){
+                self.communications(communicationsNode.map(node => {
+                    const subject = self.getNodeValue(node, 'subjects', 'subject');
+                    const subjectDescType = self.getNodeValue(node, 'subjects', 'subject description type');
+                    const type = self.getNodeValue(node, 'communication type');
+                    const date = self.getNodeValue(node, 'dates', 'date');
+                    // const endDate = self.getNodeValue(node, 'dates', 'end date');
```
aarongundel: These comments can probably be removed

arches_her/media/js/views/components/reports/consultation.js
```diff
@@ -14,31 +14,250 @@ define([
...
+                }));
+            };
+
+            const siteVisitsNode = self.getRawNodeValue(self.resource(), 'site visits');
+            if(Array.isArray(siteVisitsNode)){
+                self.siteVisits(siteVisitsNode.map(node => {
+                    const dateOfVisit = self.getNodeValue(node, 'timespan of visit', 'date of visit');
+                    const location = self.getNodeValue(node, 'location', 'location descriptions', 'location description');
+                    const locationDescType = self.getNodeValue(node, 'location', 'location descriptions', 'location description type');
+                    const attendees = self.getRawNodeValue(node, 'attendees').map(attendeeNode => {
```
aarongundel: This causes an error whenever the attendees node isn't filled in.  You should check to see if the raw node value is an array first like you do with the siteVisitsNode.  The same is true for the rest of these child nodes.

arches_her/templates/views/components/reports/consultation.htm
```diff
@@ -47,7 +44,627 @@ <h1 class="aher-report-toolbar-title"><span class="aher-report-name">{% trans "C
...
+                <a data-bind="{if: cards['site visits'], click: function(){ addNewTile(cards['site visits']) }}" class="aher-report-a">
+                    <i class="fa fa-mail-reply"></i>
+                    <span data-bind="if: cards['site visits'].tiles().length">{% trans "Edit Site Visits" %}</span>
+                    <span data-bind="ifnot: cards['site visits'].tiles().length">{% trans "Add Site Visits" %}</span>
+                </a>
+
+                <!-- Collapsible content -->
+                <div data-bind="visible: visible.siteVisits" class="aher-report-collapsible-container pad-lft">
+                    <!-- ko ifnot: siteVisits() -->
+                    <div class="aher-nodata-note">No site visit for this consultation</div>
```
aarongundel, line 633: This is never false, and doesn't show up when empty.  You may want to check for .length.  This is happening in many of the tables right now - the tables show up with "No data in table" rather than the desired message.  You may want to test by starting with an empty consultation resource

arches_her/templates/views/components/reports/consultation.htm
```diff
@@ -47,7 +44,627 @@ <h1 class="aher-report-toolbar-title"><span class="aher-report-name">{% trans "C
...
+                                <div class="aher-block-attributes">
+                                    <h4 class="aher-block-key">Date of Visit: </h4>
+                                    <div class="aher-block-value" data-bind="text: visit.dateOfVisit"></div>
+                                </div>
+                                <div class="aher-block-attributes">
+                                    <h4 class="aher-block-key">Location Description: </h4>
+                                    <div class="aher-block-value" data-bind="text: visit.location"></div>
+                                </div>
+                                <div class="aher-block-attributes">
+                                    <h4 class="aher-block-key">Attndees: </h4>
```
aarongundel: Typo

arches_her/templates/views/components/reports/consultation.htm
```diff
@@ -47,7 +44,627 @@ <h1 class="aher-report-toolbar-title"><span class="aher-report-name">{% trans "C
...
+                            </div>
+                        </div>
+                    </div>
+                    <!-- /ko -->
+                </div>
+            </div>
+            <!-- Consultation Progression Tab -->
+            <div class="aher-report-page" data-bind="if: activeSection() === 'progression'">
+                <div class="aher-report-section">
+                    <h2 class="aher-report-section-title"><i data-bind="click: function() {toggleVisibility(visible.proposal)}, css: {'fa-angle-double-right': visible.proposal(), 'fa-angle-double-up': !visible.proposal()}"  class="fa toggle"></i> {% trans "proposal" %}</h2>
```
aarongundel: capitalize proposal

arches_her/templates/views/components/reports/consultation.htm
```diff
@@ -47,7 +44,627 @@ <h1 class="aher-report-toolbar-title"><span class="aher-report-name">{% trans "C
...
+                                                <th>{% trans "Reference Type" %}</th>
+                                                <th>{% trans "Agency" %}</th>
+                                                <th class="aher-table-control all"></th>
+                                            </tr>
+                                        </thead>
+                                        <tbody data-bind="dataTablesForEach: { data: systemReferences, dataTableOptions: createTableConfig(4)}">
+                                            <tr>
+                                                <td data-bind="text: reference"></td>
+                                                <td data-bind="text: referenceType"></td>
+                                                <td data-bind="text: agency"></td>
```
aarongundel: agency should be a hyperlink to another resource.

arches_her/templates/views/components/reports/consultation.htm
```diff
@@ -47,7 +44,627 @@ <h1 class="aher-report-toolbar-title"><span class="aher-report-name">{% trans "C
...
+                <!-- Collapsible content -->
+                <div data-bind="visible: visible.contacts" class="aher-report-collapsible-container pad-lft">
+                    <!-- ko ifnot: contacts() -->
+                    <div class="aher-nodata-note">No contact for this consultation</div>
+                    <!-- /ko -->
+
+                    <!-- ko if: contacts() -->
+                    <div class="aher-report-subsection">
+
+                        <div class="firstchild-container">
```
aarongundel, line 190: This might be better served with the keyvalue scene that already handles this kind of data - but these should be resource links.  Also there is no consulting contact card visible in the editor, so this will always be blank.

aarongundel (changes requested):
Looks pretty good!  Couple of small changes.

arches_her/media/js/views/components/reports/consultation.js
njkim: Yes, thanks.

arches_her/templates/views/components/reports/consultation.htm
njkim: good catch!

arches_her/media/js/views/components/reports/consultation.js
njkim: Updated, thanks.

arches_her/templates/views/components/reports/consultation.htm
njkim: Fixed and a couple of other places with the similar if statements.

arches_her/templates/views/components/reports/consultation.htm
njkim: :)

arches_her/templates/views/components/reports/consultation.htm
njkim: Updated

arches_her/templates/views/components/reports/consultation.htm
njkim: Some of the nodes are the resource-list datatype so the nodes are updated to represented as tables.
As a result, ended up making it more complex custom tab, instead of using default scene.

aarongundel:
looks good!

aarongundel (approved):

PR #729, opened 2021-12-06
Adds new reports

aarongundel:
This adds new reports in the style of Arches for Science.


chiatt (approved):
These reports look really good. I noticed a couple of minor issues (e.g. geometries not showing up in Heritage Story), but these can be addressed as separate issues once this is merged in.

Issue #697, opened 2021-08-17
Removal of the Consultations Find page


rlwilliamss:
**Description** -  Logging this ticket as a placeholder for the removal of the Consultations Find page, leaving just the standard search page as the main searching tool.

![image](https://user-images.githubusercontent.com/78143228/129738280-3d7a4430-b979-4c7a-9aa5-0e4448079f72.png)

**Index page** 

Provide the following links:
- Consultations (drop down menu) 
      - Active Consultations 
      - New Workflow 

- “Search” (to the normal Arches search) 
- “Manage” (to standard Arches manage resources) 
- Remove “Graph” 

![menu](https://user-images.githubusercontent.com/78143228/129739193-92f645d0-8866-49a7-9b89-216b4fc4dfbf.png)

- Remove references to GLAAS as this will be a generic tool.

**Custom Consultations interface**

- Move the “Active Consultations” and “New” button/pages to the existing navigation bar:

![longmenu](https://user-images.githubusercontent.com/78143228/129739152-91bf9ec2-f373-41e3-b9c2-82b6ef5d8b2b.png)

- Remove the rest of the custom consultations module UI.

chrisdolby-HE:
@dwuthrich @rlwilliamss Please note that this hasn't in fact been decided. This should not be an issue, were the new proposed report templates to be approved and adopted. Were this not to be the case, we would need to confirm this with Sarah Harrison.

chiatt:
Resolved in #657

PR #730, opened 2021-12-09
Remove find page and other customizations

chiatt:
Removes the find page and the alternate navigation menu, re #697


njkim (approved):
All relevant pages are function fine.

PR #732, opened 2021-12-10
Fix the issue showing the reference number instead of type, #683

### LINKED ISSUE #683: Consultation workflow - Reference Type column displays Reference Numbers
**Issue Description:** (priority 1 bug) - When on the summary page, the reference type column is displaying the reference numbers as shown in the reference number column, rather than the reference type:

image

njkim:
Fix the final step showing reference number instead of reference type, #683


chiatt (approved):
Looks good to me.

PR #735, opened 2021-12-16
update the django view to format letter, #675

### LINKED ISSUE #675: Consultation letters - empty fields display as 'none'
**Issue Description** -  (priority 1 - enhancement) - On the Consultation letters, if a notes field is left empty (i.e. for the Mitigation) instead of this section being left blank on the letter, the header ‘Mitigation’ remains. 

image

Additionally, in the Address section of the letter, if a field is left blank, the word ‘none’ appears rather than nothing appearing.

image

njkim:
To fix the issue of replacing the value with "None" or not replacing the value at all if the value does not exist, #675


chiatt (approved):
Works as expected

Issue #738, opened 2021-12-16
Consultation workflow: Proposal description does not show in the final step.


chiatt:
The proposal description does not show in the final step.

chiatt:
@njkim Related Files also appears not to be updated in the final step

chiatt:
Fixed in #741

PR #741, opened 2021-12-16
Update consultation final step, #738

njkim:
Update the final step of the consultation step to show the proposal and the related file properly, #738


chiatt (approved):
Looks great

PR #743, opened 2021-12-20
Improve ms docx letter formatting, #676

### LINKED ISSUE #630: Remove editing comments from correspondence templates
If a user completes the correspondence workflow and then opens the resulting Word docx file, they will be able to see editor comments from people that worked on the templates. This information should be cleared so that the document edit history is clean.

### LINKED ISSUE #676: Consultation letters - format, font and text size should be consistent
**Issue Description** - (priority 1 – bug): The letters are compiled using a range of fonts/text sizes. The sections that are pulled in from the consultation record are in a different font. The font we would like to use is Source Sans Pro, and the font size as 12 as this is what the current letters use. 

image

Some letters are printing with strange formatting, e.g. [OL2] tags. Is this because of the Window view settings (protected view?) 
We need to have the comments removed; we have been made aware that there are some forthcoming changes but we are not in a position to put them in place as we have not been provided with the modifications to the templates.

image

njkim:
The comments are removed, #630 
The default font is set to Source Sans Pro 12 pt, #676


chiatt (approved):
These look much better!

PR #745, opened 2021-12-21
pep 517 is causing problems for the install process

aarongundel:
There's a problem with editable mode in the current arches her container build.  Skirting pep 517 is the only way we've found to workaround the problem.  There's probably a larger fix that'll need to be made to make us compatible with pep 517.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,7 +59,7 @@ COPY ./arches ${ARCHES_ROOT}
 # From here, run commands from ARCHES_ROOT
 WORKDIR ${ARCHES_ROOT}
 
-RUN pip install -e . && pip install -r arches/install/requirements_dev.txt
+RUN pip install -e . --no-use-pep517 && pip install -r arches/install/requirements_dev.txt
 
 COPY /arches_her/docker/entrypoint.sh ${WEB_ROOT}/entrypoint.sh
 RUN chmod -R 700 ${WEB_ROOT}/entrypoint.sh &&\


chiatt (approved):
Works as expected

aj-he:
@chiatt @aarongundel  This has not fixed the issue for me. I needed to add `--user` flag to for the module to be properly registered. The full working line for me is as follows:

`ln 62: RUN pip install --user --no-use-pep517 -e . && pip install -r arches/install/requirements_dev.txt`

My environment is Windows 10 using docker in WSL2.

This makes me think there is a permissions issue at play also.  Perhaps we need to bring back a venv to help with this?

aarongundel:
@aj-he Thanks for testing that.  We were able to get it (the original solution) working for some - but not all - of us.  I'm wondering if the issue is, at the core, the build system for arches.  I'm doing some research on that now - as in the long term, we wouldn't want to run with --no-use-pep517 anyway (pep517/518 are our friends).  

It's possible we may have to bring back a venv, but that may be masking the core problem and adds another layer of abstraction to the container.  On the other hand, it gives us more control over the python versioning.  Alternatively, we may need to switch container bases to something where we have more control over the distribution we're using.  I'm going to do some research over the long weekend and circle back next week on this.  Thanks for sharing that --user flag, we're testing that out now here.

aj-he:
@aarongundel did you get anywhere on this? I am going to get my team to use the `--user` flag as we are having issues without it.

PR #756, opened 2021-12-28
Do not show the hidden cards (audit), #751

### LINKED ISSUE #751: Report templates show data for hidden cards
The HER templates show information for hidden cards such as in the "Audit Data" tab.  Hidden cards should not appear in the report template

njkim:
Hide the hidden card (audit...) in consultation custom report, #751

### DIFF
--- a/arches_her/media/js/views/components/reports/activity.js
+++ b/arches_her/media/js/views/components/reports/activity.js
@@ -65,7 +65,6 @@ define([
             }
 
             self.nameCards = {};
-            self.auditCards = {};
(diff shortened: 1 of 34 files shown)


njkim:
It appears that `audit` is hidden in many other graphs too (not just consultations)

chiatt (approved):
Looks good to me. This resolves the problem for the audit card, however the larger problem of getting 'hidden' tile data back from the resource json endpoint still needs to be addressed: archesproject/arches#7838

Issue #750, opened 2021-12-21
Consultation Report template is broken


dwuthrich:
There is a problem with the "contacts" tab.  After pressing it, all subsequent tabs appear below the contacts instead of within their own tab

PR #757, opened 2021-12-28
Fix dataTable contents issue, #750

njkim:
Make the datatable contents an observable for the recent update, #750


chiatt (approved):
Looks good to me!

PR #761, opened 2022-01-04
Fix not loading designation/protection if no geometry, #759

### LINKED ISSUE #759: Designation / Protection tab does not load if geometry is not available in the custom report
For example, in the heritage asset / area and other resource

### LINKED ISSUE #765: Land Use tab is not showing the values
In several custom reports
eg. heritage asset, historic aircraft, activity, etc

njkim:
Check features is available before creating featureCollection, #759 
Fix the land use value not appearing by take the n-card into account, #765 
Also fix a minor issue with historic aircraft not showing ownership properly.

### DIFF
--- a/arches_her/media/js/views/components/reports/artefact.js
+++ b/arches_her/media/js/views/components/reports/artefact.js
@@ -62,6 +62,7 @@ define([
             };
 
             self.protectionDataConfig = {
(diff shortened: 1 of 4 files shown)


chiatt (approved):
Looks good to me

PR #769, opened 2022-01-05
The missing values in the report, #768

### LINKED ISSUE #752: Consultation report fails to load correctly
Loading a consultation report will fail and the console shows a reference to JSON not correctly parsing on the following line.

```js
const url = JSON.parse(self.getNodeValue(node, 'url')).url;
```

inspection shows that the value returned by `self.getNodeValue(node, 'url')` is `"--"` so this needs to be caught.

### LINKED ISSUE #768: Some values are not displayed in the reports
1. consultations
url, url label
site visit (observed by, recommended by)

2. heritage asset
associated activities

3. heritage story
location data
administrative area

4. historic aircraft
journey details has 2 flights

njkim:
Update various reports to add missing values, #768

which include
consultations - site visit (observed by, recommended by)
heritage asset - associated activities
heritage story - location data, administrative area
historic aircraft - last flight

Also fix cross ref url & url label (#752) and some missing card issues in consultation


chiatt (approved):
Looks good to me

Issue #775, opened 2022-01-05
Names Scene: Scene shows table even when instance has no names


dwuthrich:
As shown in the image, the name scene shows an empty table instead of a "no names for this resource" message.

What should happen?  The scene should only show a message (not an empty table) when there are no names associated with a resource![Screen Shot 2022-01-05 at 10.26.17 AM.png](https://images.zenhubusercontent.com/56f1ad0a6caf56c9d38e1cbd/49d55e14-5be5-4904-8170-a6614faa2fab)

chiatt:
Fixed in #778

PR #778, opened 2022-01-06
Not showing a name table when there is no name, #775

njkim:
By checking if the name is available in the name tab, #775


chiatt (approved):
lgtm!

Issue #780, opened 2022-01-06
Edit buttons are missing in report in editor


njkim:
eg. Activity
 - classification and dating
 - associate resources

PR #782, opened 2022-01-07
Missing edit button due to missing cards in activity report, #780

njkim:
activity type
activity timespan


chiatt (changes requested):
The edit buttons are now available, but I can't edit a Type. If I click the edit type button it loads an empty form to create a new type tile.

image

chiatt (approved):
Looks good!

Issue #791, opened 2022-01-14
dateFormat config property in EDTF nodes for Maritime Model


aj-he:
There are two **edtf** nodes in the maritime graph that seem to have retained a config property from when they used to be a **date**. This affects **Construction Phase End Date** and **Construction Phase Start Date**.

Property must be removed.

```json
{
        "config": {
                "dateFormat": "YYYY-MM-DD", <<<<<<<<<<<<<
                "fuzzy_day_padding": 1,
                "fuzzy_month_padding": 1,
                "fuzzy_season_padding": 12,
                "fuzzy_year_padding": 1,
                "multiplier_if_approximate": 1,
                "multiplier_if_both": 1,
                "multiplier_if_uncertain": 1
        },
        "datatype": "edtf",
        "description": null,
        "exportable": true,
        "fieldname": "End_date",
        "graph_id": "49bac32e-5464-11e9-a6e2-000d3ab1e588",
        "is_collector": false,
        "isrequired": false,
        "issearchable": true,
        "istopnode": false,
        "name": "Construction Phase End Date",
        "nodegroup_id": "656c41a9-3ec0-11eb-a444-f875a44e0e11",
        "nodeid": "0557783e-535d-11ec-b437-a87eeabdefba",
        "ontologyclass": "http://www.cidoc-crm.org/cidoc-crm/E61_Time_Primitive",
        "parentproperty": "http://www.cidoc-crm.org/cidoc-crm/P82b_end_of_the_end",
        "sortorder": 0
},
{
        "config": {
                "dateFormat": "YYYY-MM-DD", <<<<<<<<<<<
                "fuzzy_day_padding": 1,
                "fuzzy_month_padding": 1,
                "fuzzy_season_padding": 12,
                "fuzzy_year_padding": 1,
                "multiplier_if_approximate": 1,
                "multiplier_if_both": 1,
                "multiplier_if_uncertain": 1
        },
        "datatype": "edtf",
        "description": null,
        "exportable": true,
        "fieldname": "Start_date",
        "graph_id": "49bac32e-5464-11e9-a6e2-000d3ab1e588",
        "is_collector": false,
        "isrequired": false,
        "issearchable": true,
        "istopnode": false,
        "name": "Construction Phase Start Date",
        "nodegroup_id": "656c41a9-3ec0-11eb-a444-f875a44e0e11",
        "nodeid": "dcd1a180-535c-11ec-8227-a87eeabdefba",
[...]

PR #792, opened 2022-01-18
removed incorrect dateFormat property in EDTF nodes

aj-he:
Addresses #791 

Removed two instances where EDTF nodes in Maritime Vessel resource model graph had a dateFormat property.


chiatt (approved):
Looks good!

PR #880, opened 2022-02-18
719 add generic set of html export templates

RobOatesHistoricEngland:
Added 9 export templates each linking to a custom header file and a custom style file (both .htm).  The templates are for:

- Consultation
- Application Area
- Maritime Craft
- Heritage Asset
- Place
- Historic Landscape Characterization
- Heritage Area
- Historic Aircraft
- Activity


chiatt (approved):
These all look good to me

Issue #855, opened 2022-02-03
Correspondence workflow letters - H Condition Satisfied casework officer phone/email not displaying


rlwilliamss:
**Describe the issue** - In letter template H - Condition Satisfied , Casework Officer Email and Direct Dial lookup are not displaying:

![image](https://user-images.githubusercontent.com/78143228/152397309-2c46164e-ef03-4adf-a47a-effa076400d9.png)

chiatt:
Resolved in #890

Issue #854, opened 2022-02-03
Correspondence workflow letters - Letter G PXA Approval not found


rlwilliamss:
**Describe the issue** - The Template for letter G – PXA Approval is not being found so this workflow can’t be initiated. The template is there, just with the full version of the (admittedly unindicative) acronym – PXA = Post Excavation Assessment. The template file in question is therefore present in keystone --> arches-her --> arches_her --> docx --> Post Excavation Assessment Approval Letter.docx.

![image](https://user-images.githubusercontent.com/78143228/152396687-57112c21-24e6-4e23-aa33-120cd447bee4.png)

Issue #853, opened 2022-02-03
Correspondence workflow letters - F1 WSI Approval casework officer email not displaying


rlwilliamss:
**Describe the issue** - In letter template F1 – WSI Approval, Casework Officer Email lookup is not working:

![image](https://user-images.githubusercontent.com/78143228/152394328-968d0f7c-2090-4c63-8325-d683dd74f25c.png)

chiatt:
Resolved in #890

Issue #852, opened 2022-02-03
Correspondence workflow letters - F1/F2 address text colour in blue


rlwilliamss:
**Describe the issue** - in letter templates F1 – WSI Approval and F2 – WSI Amend, the whole address has been formatted in blue, rather than just the contact’s name. Not sure which is correct but needs to be made consistent.

Also, text being pulled from the model in template letter type B1 - Pre app predetermination EAS version is not being included in bold type. See Desk Based Assessment text below.

![image](https://user-images.githubusercontent.com/78143228/152392760-0fd230dd-3186-4e6e-81fe-56f9f7a8dad0.png)

![image](https://user-images.githubusercontent.com/78143228/152393547-c2bac371-897f-4bb8-9a71-83f45f2463ea.png)

rlwilliamss:
Update - All text in the letters should display in black.

chiatt:
Resolved in #890

PR #890, opened 2022-03-25
Fix letter templates

njkim:
Update correspondence step and the letter templates

make the correspondence letter step required
add a template and link the template & concept value, #785
update heritage asset/area logic, #851 
update template to make all the text color black, #852
link the template Letter G PXA Approval, #854
fix key words in a various templates, #853, #855


chiatt (approved):
#785, #851, #852, #853, #855 all look good.

Issue #879, opened 2022-02-17
Application Area workflow - add multiple Designation and Protection Assignments in Area Designations step


rlwilliamss:
**Describe the issue** - Can the Area Designations step in the Appication Area workflow be changed so that we can add multiple Designation and Protection Assignments through the workflow, as the Designation and Protection Assignment card is now a multi select widget.

**Steps to reproduce**

1) Open a new Application Area workflow and fill out the steps to the Area Designations step

2) At the moment you can only add one Designation and Protection Assignment

![image](https://user-images.githubusercontent.com/78143228/154484042-ab3ea7e7-953e-4bed-918e-a8dd45852329.png)

PR #896, opened 2022-03-31
Make area designation step multi-tile, #879

njkim:
Update app area workflow area designation step to a multi-tile step, #879

### DIFF
--- a/arches_her/media/js/views/components/plugins/application-area.js
+++ b/arches_her/media/js/views/components/plugins/application-area.js
@@ -127,9 +127,9 @@ define([
                         {
                             componentConfigs: [
                                 { 
-                                    componentName: 'map-card',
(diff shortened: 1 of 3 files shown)


chiatt (changes requested):
This looks good @njkim, but the final step shows only the first of the designations if there is more than one.

njkim:
> the final step shows only the first of the designations if there is more than one.

Ah, good catch! thanks. I am looking into it.

njkim:
> This looks good @njkim, but the final step shows only the first of the designations if there is more than one.

The summary step is updated. Also, all other visible nodes in the summary are added, too.

chiatt (approved):

chiatt (approved):
Looks great!

PR #900, opened 2022-04-09
Display alert when required node has no data, re #899

### LINKED ISSUE #899: Communication Workflow - Notes step not loading
**Describe the issue** - Communication Workflow - the Notes step is not loading.

**Steps to reproduce**

1) Open a new Communication workflow

2) Select a Consultation from the drop down, enter data into the rest of the fields and click save and continue to complete the first step of the worfklow

3) The Notes step does not load - if you refresh the page you are taken back to step 1

image

chiatt:
Fixes bug causing Communication workflow to stall on first step by dropping the loading spinner and displaying an alert when required node has no data, re #899

### DIFF
--- a/arches_her/media/js/views/components/workflows/communication-workflow/communication-select-resource.js
+++ b/arches_her/media/js/views/components/workflows/communication-workflow/communication-select-resource.js
@@ -3,8 +3,8 @@ define([
     'knockout',
     'uuid',
     'arches',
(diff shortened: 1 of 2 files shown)


njkim (changes requested):
I believe you plan to indicate (ie. * ) that the node value is required.

njkim (approved):
works great

Issue #921, opened 2022-05-04
Active consultation card maps excess data loading


aj-he:
Each card in the active consultations plug-in has its own map and it appears that for each map it downloads the three geojson map sources for selecting related areas.

This will affect loading performance and download a lot of data as the number of active consultations increases.

aj-he:
@chiatt this happens for all maps, the three sources download the geojson each time. Active consultations is noticeably affected but it would be good if the sources only used for related resources map card were only added to the map then to save time/data

aj-he:
@chiatt I've not closed this as PR #932 was a temporary measure.

PR #932, opened 2022-05-07
Remove unused map sources, re #921

chiatt:
This is a stop-gap measure to remove unused layers that result in calls to the geojson endpoint each time the map component is loaded. This change will prevent making 2 of the 3 requests referred to in #921. A complete solution would be to also replace the `select-application-area` with a vector tile source, eliminating all of the unnecessary requests.

The changes in this PR will only be effective when the package is loaded into a new database. 

To apply this to an existing HER project without reloading the package, you can run:
```
delete from map_sources where name in ('select-heritage-asset','select-heritage-area');
```


aj-he (approved):
this works for me

Issue #930, opened 2022-05-06
HTML custom reports - Parent Asset not displaying in Heritage Asset report


rlwilliamss:
**Issue Description** - In the Heritage Asset HTML report, the Parent Assets are not displaying.

**Steps to reproduce**

1)  Search for a Heritage Asset record/s with Parent Asset data

2) Click the Search Export button, select the HTML option and download the export

3) Open the HTML report - the Parent Assets are not dispaying in the report e.g., 

![image](https://user-images.githubusercontent.com/78143228/167135763-17a8810c-c6fd-49ca-af94-0f9d2104148d.png)

![image](https://user-images.githubusercontent.com/78143228/167135787-9f078a51-11b8-40c1-baad-a99c8d7c933e.png)

aj-he:
@csmith-he I've assigned this to you as you have picked this up internally

chiatt:
Resolved in #933

PR #933, opened 2022-05-09
930 HTML report - Parent Asset fix

csmith-he:
The HTML report issue #930 has been fixed so the Parent Asset now displayed in the HTML export report.


chiatt (approved):
Looks good to me!

Issue #929, opened 2022-05-06
HTML custom reports - Associated Resources not displaying


rlwilliamss:
**Issue Decription** - Related/Associated Resources are not displaying in the custom HTML reports.

**Steps to reproduce**

1) Search for a Consultation record/s with Associated Resources

2) Click the Search Export button, select the HTML option and download the export 

3) Open the HTML report - Associated Resources are not displaying

PR #934, opened 2022-05-09
HTML report - Associated Resources fix

csmith-he:
Pull after #933 

The HTML report issue #929  has been fixed so the Associated Resources now displayed in the HTML export report.


chiatt (approved):
These changes look good to me.

Issue #868, opened 2022-02-11
Correspondence letters - Address is linked to Person


rlwilliamss:
**Describe the issue** - Address is linked to Person.  This seems to an issue with some addresses where the organisation has a name as well. 

Example:
London Borough of Tower Hamlets, Town Hall, Mulberry Place, 5 Clove Crescent, London E12 2BG

Try to fill out the correct fields:
![image](https://user-images.githubusercontent.com/78143228/153566347-d874325b-76ca-4f07-9382-54352c150871.png)

This is the letter:
![image](https://user-images.githubusercontent.com/78143228/153566432-f4cacd5f-e827-4cb8-b6d6-d854d4e1564c.png)

To get the below I have to put London Borough of Tower Hamlets, Town Hall, Mulberry Place all in the ‘Building Name’ field:
![image](https://user-images.githubusercontent.com/78143228/153566498-613ae85c-368a-4b72-bf66-91b84ffb23b7.png)

Also, the Person Title is missing, can this be added to letters.

![image](https://user-images.githubusercontent.com/78143228/153567678-09ffdc15-f356-424f-918a-23dcf27ca6cb.png)

Issue #850, opened 2022-02-03
Correspondence workflow letters - update Address


rlwilliamss:
**Describe the issue** - Currently, the letter address is looking up Person -> Address -> Full Address, and the term of address is looking up to Person -> Name -> Full Name. 

Would it be possible to redirect these to look at Person -> Contact Details (where Contact Point Type = Mail) -> Contact point for the letter address. 

And Person -> Contact Details (where Contact Point Type = Mail) -> Contact Name for Correspondence for the Name, so that 'Dear Charlie O'Sullivan' would be 'Dear Mr O'Sullivan'.

![image](https://user-images.githubusercontent.com/78143228/152387713-8f710d75-97c2-4659-a784-e77c4c2bbc15.png)

chiatt:
The contact point is a single line of text. It may not always be possible to parse it into a formatted address.

dwuthrich:
We could make the string widget for the address node in the person/organization model a rich-text widget.  That would allow the editor to format the address into multiple lines

chiatt:
@dwuthrich That's a good suggestion. Unfortunately we can't insert html into the Word document, but we could at least use the html to figure out where the line breaks are and then strip the html out of the string.

PR #935, opened 2022-05-09
Update the source of the correspondence address, #850

njkim:
Update the source of receiver's address from person'e address to contact details, #850 
Also add the name title, #868 

"If" the rich-text widget is used for the node, the address will be appears formatted. Otherwise it will appear in a single line.


chiatt (approved):
Looks good to me!

PR #941, opened 2022-05-12
Prevent error in consultations final step when consultation has no geom

### LINKED ISSUE #940: If a consultation is not assigned a geometry, the final step of the consultation workflow fails
If a consultation is not assigned a geometry, the final step of the consultation workflow fails. If a consultation does not have a geometry, the final step should load as expected, and the map should not be rendered.

chiatt:
Prevents an error in consultations final step when consultation has no geometry. Also, does not show map if there is no geometry to present.


aj-he (approved):
works well

PR #947, opened 2022-05-17
Make the digital object filename a clickable link

### LINKED ISSUE #794: How to change the Digital Object to be able to download file
Hi,
We have a client who wants to use arches (arches-her) project.
I'm new to arches and I'm experimenting with the resource models trying to adjust them according to the clients requests.
We want to change the Digital object model to allow download of the file/s.
I cannot find a way how to achieve this.
Looking at the V5 demo I see the Information Source type, which allows download, but it is using a template called Image Header Template.
Switching the template from digital object in "her" to Image Header Template the file can be downloaded.
But I would like to have the same download feature in Digital Object template, because it is much better for the purpose.
I see the implementation of Digital Object Template but I cannot find how the download is implemented in Image Header template.
Any advice how to do this is highly appreciated.

### LINKED ISSUE #917: Reports: Digital Object - unable to download file from report
**Issue Description** - In the Digital Object resource report, the file under the File Details tab is not clickable as a link to download the file.

**Steps to reproduce**

1) Open a Digital Object in report mode

2) Click on the File Details tab

3) The File value is not currently a downloadable link - if possible users would like this to be a link which downloads the file

image

phudson-he:
In the Digital Object resource report, make the file under the File Details tab clickable as a link to download the file. #917

### DIFF
--- a/arches_her/media/js/views/components/reports/digital-object.js
+++ b/arches_her/media/js/views/components/reports/digital-object.js
@@ -36,6 +36,17 @@ define([
             self.cards = {};
             self.copyrightCards = {};
 
+            self.visible = {
+                files: ko.observable(true),
+            }
+
+            self.createTableConfig = function(col) {
+                return {
+                    ...self.defaultTableConfig,
+                    columns: Array(col).fill(null)
+                };
+            }
+
             if(params.report.cards){
                 const cards = params.report.cards;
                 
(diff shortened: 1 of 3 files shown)


arches_her/media/js/views/components/reports/digital-object.js
```diff
@@ -64,7 +64,8 @@ define([
                             data: [{
                                 key: 'File',
                                 value: self.getFileName(self.getRawNodeValue(self.resource(), 'file content', 'file')),
-                                type: 'kv',
+                                href: self.getNodeValue(self.resource(), 'file content', 'file'),
```
aj-he: This seems to be creating a bad link address. In this example the link should be *http://localhost:8002/files/d0e52338-cd3c-11ec-872b-0242ac130007* but it is appearing as *http://localhost:8002/files/d0e52338-cd3c-11ec-872b-0242ac130007%20%7C%20/files/d0e9e4ae-cd3c-11ec-872b-0242ac130007*

image

aj-he: ... and what I can see from the Dev Tools...
image

aj-he (changes requested):
See the comments on the file link not forming correctly.

arches_her/media/js/views/components/reports/digital-object.js
phudson-he: Now enabled multiple files to display as expected in report view/editor

aj-he (approved):

aj-he:
fixes #794 also

Issue #948, opened 2022-05-17
Need to add context param to the function save and post_save


aj-he:
A recent change to arches/dev/6.1.x where tile saves pass context to functions, means we need to update functions in arches-her

chiatt:
Resolved in #949

PR #949, opened 2022-05-17
fixes func save params #948

aj-he:
adds the new context param required for function save method in dev/6.1.x


chiatt (approved):
Looks good to me.

chiatt (commented):
In hindsight I think a safer solution would be to pass in kwargs, like so:

```def save(self, tile, request, **kwargs):```

That way the function is safe from the addition of any future optional arguments.

PR #958, opened 2022-05-20
Updates application area with Name and Location Description steps, re #939 

### LINKED ISSUE #939: Add Location Descriptions to the Assign Address step in Application Area workflow
For Application Area workflow,
add the Location Descriptions card to the Assign Address step
so that it can be used to record the location were there is no postal address.

The Application Area Name will need to be filled manually.

chiatt:
Updates the application area workflow with Name and Location Description steps. Removes the address step, re #939

### DIFF
--- a/arches_her/media/js/views/components/plugins/application-area.js
+++ b/arches_her/media/js/views/components/plugins/application-area.js
@@ -11,25 +11,23 @@ define([
             this.componentName = 'application-area';
             this.stepConfig = [
                 {
-                    title: 'Assign Address',
-                    name: 'assign-address',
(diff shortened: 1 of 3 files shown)


njkim (changes requested):
@chiatt Overall workflows is working great.
But the "Area Designation" step (multi-tile step) does not show the add button. Also, the summary step does not seem to show all the contents at the bottom.
A height adjustment may be necessary?

njkim (approved):
The workflow works great.
The css issue does not appear in the 6.1.x. So, all looks good.

PR #977, opened 2022-05-24
Changed Classifications and Dating to Source Type

### LINKED ISSUE #943: Bibliographic Source report - rename Classifications and Dating as there is no dating
In the Bibliographic Source report template, could the Classifications and Dating tab be renamed as there is no dating. It would be better called **Bibliographic Source Types**.

image

phudson-he:
In the Bibliographic Source report template, changed the "Classifications and Dating" tab be "Source Type".


aj-he (approved):
all fine

PR #984, opened 2022-05-25
Made Period and Historic Map links clickable

phudson-he:
Historic Landscape Characterization (HLC) model - ensure that link text that links to associated resources in the report template is rendering href links as expected, and also ensure that multiple links are also displaying/rendering correctly instead of a single entity.

Includes fix for HLC phase description also not loading.


aj-he (approved):
works well for me

PR #987, opened 2022-06-07
Add ability to query BNG in advanced search

### LINKED ISSUE #986: BNG Datatype has no datatype config files
**Issue description** - A node with datatype set to bngcentrepoint does not have the option to add the node to the Advanced Search.

It was determined that the switch for adding the datatype to Advanced Search as not appearing because there was no configname stored for the datatype in the d_data_types table as it was not specified in the datatype config.

image

Once a value is given for config name, the switch appears and National Grid Reference appears in the Advanced Search options.  However, without the configuration files (.js and htm) within the /views/components/datatypes folders for media and templates, the user is not able to search on national grid references as there is no useable UI.

To complete this ticket the following needs to be completed:

- configcomponent and configname defined in the bngcentrepoint.py file
- configuration component (.js and .htm) created for bngcentrepoint datatype (for more details: https://arches.readthedocs.io/en/stable/creating-custom-components/#datatypes)

csmith-he:
The BNG Datatype didn't have configuration files to allow it to be queried in the Advanced Search.

Fixes #986

### DIFF
--- a/arches_her/datatypes/bngcentrepoint.py
+++ b/arches_her/datatypes/bngcentrepoint.py
@@ -1,6 +1,8 @@
 from arches.app.datatypes.base import BaseDataType
 from arches.app.models.models import Widget
 from arches.app.models.system_settings import settings
+from arches.app.search.elasticsearch_dsl_builder import Match, Exists
+import logging
 
 bngpoint = Widget.objects.get(name="bngpoint")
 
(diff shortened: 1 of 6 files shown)


aj-he:
NOTE: need to run the following against an existing install register the updates to BNG datatype components

`python3 manage.py datatype update --source /web_root/arches_her/arches_her/datatypes/bngcentrepoint.py`

chiatt (approved):
This looks great @csmith-he. One consideration is that the validate method in core arches takes a `**kwargs` argument. We weren't really using that until the latest in 6.1.x, but now it is required in datatype validate methods. You'll need to update that when you pull the latest from arches dev/6.1.x. However, if you wanted to make that change now, we could merge it in with this PR. You also could do it in a later PR if you prefer. The change is pretty simple. You'd just need to add the `**kwargs` arg on line 24 of `bngcentrepoint.py` like so:

```
class BNGCentreDataType(BaseDataType):
    def validate(self, value, row_number=None, source=None, node=None, nodeid=None, strict=False, **kwargs):
```

aj-he:
@csmith-he can you apply the required changes suggested by @chiatt to get this completed please.

PR #993, opened 2022-07-15
Update graphs with alias changes and remove cruft

chiatt:
Updates the Aircraft resource model with aliases that correspond to recent changes in node names.


njkim (approved):
Looks good.

PR #994, opened 2022-07-17
Update graphs with hasnodealias value

chiatt:
Update resource models with hasnodealias and historic aircraft with alias changes, re archesproject/arches#8220


njkim (approved):
Looks good

PR #997, opened 2022-07-20
Update template dependency for ver 7.0

njkim:
Update template dependencies in viewmodels.


njkim:
The pr is replaced by more extensive change  and thus closed.
#1000

Issue #995, opened 2022-07-18
Archive Holding values not displaying


csmith-he:
Archive Holding data is not displaying in Archive Source report.

Steps to reproduce

1) Open Archive Source with Archive Holding populated.

2) Click on the Archive Holding report tab - Archive Holding data is not displaying

![image](https://user-images.githubusercontent.com/38533555/179531973-5864c04f-e3a2-4c01-9278-dc6fb0d124e2.png)

Issue #988, opened 2022-06-09
Custom HTML reports - single Activity Type data not displaying


rlwilliamss:
**Issue description** - Activity records -  Activity Type data not displaying in HTML reports where there is only one Activity Type.

**Steps to reproduce**

1) Open the search page

2) Click on the Resource Type drop down and select Activity

3) Draw a small polygon on the map to limit the results to 20 or so records

4) Click on the search export button and select the HTML option

5) Download the HTML report and open it - Activity Type data is not displaying for records with only one Activity Type

![image](https://user-images.githubusercontent.com/78143228/172832767-9c1b9aea-8cb4-4bf0-ba54-cc7f4971cd38.png)

![image](https://user-images.githubusercontent.com/78143228/172832791-93e5690c-065a-4745-93ba-436840f5e132.png)

Issue #905, opened 2022-04-12
Rename Heritage Asset and Heritage Area graphs


aj-he:
(To be signed off)

The Heritage Asset and Heritage Area models are using a term that has a different meaning/legal status and could be confusing and therefore needs changing.

The new model names will reflect the alternative terms referring to these resource instances

"Heritage Asset" to become "Monument"
"Heritage Area" to become "Area"

aj-he:
fixed by #1003

PR #1003, opened 2022-07-26
905 988 995 Graph Rename and Other Fixes

csmith-he:
Fixes for tickets #905, #988 and #995


chiatt (approved):
I think everything looks good here. Reports, workflows work as expected.

PR #1004, opened 2022-07-26
updates to docker build process

aarongundel:
adds settings_local for easy install, gets pip from apt, adds requirements, adds python-is-python3 for ease of use, makes sure yarn builds after run because volumes


chiatt (approved):
Looks good!

PR #1007, opened 2022-08-10
Adds docker webpack configuration

aarongundel:
Adds a new docker webpack configuration.  Makes the build server a side container that runs in the background for easier logging and restarting when configuration changes.


chiatt (approved):
Looks good!

PR #1013, opened 2022-08-17
1012 cbyrd webapck updates

chrabyrd:
General webpack Quality of Life updates: explicitly sets 8GB memory in webpack processes to prevent heaperror, colors warnings/errors of build process, and handles for errant .DS_Store files


robgaston (approved):
looks good

PR #1024, opened 2022-09-14
Update docker for 7.1

aarongundel:
Changes for docker 7.1 compatibility.  Adds python to webpack machine since settings have migrated to settings.py


arches_her/webpack/webpack.common.js
```diff
@@ -243,6 +239,19 @@ module.exports = () => {
                     ],
                 },
             });
+        };
+
+        let projectSettings = spawn(
```
chrabyrd, line 244: ah -- this pattern has been updated since we last spoke

```
        let projectSettings = spawn(
            'python3',
            [Path.resolve(__dirname, Path.parse(__dirname)['dir'], 'settings.py')]
        );
        projectSettings.stderr.on("data", process.stderr.write);
        projectSettings.stdout.on("data", createWebpackConfig);

        projectSettings.on('error', () => {
            projectSettings = spawn(
                'python',
                [Path.resolve(__dirname, Path.parse(__dirname)['dir'], 'settings.py')]
            );
            projectSettings.stderr.on("data", process.stderr.write);
            projectSettings.stdout.on("data", createWebpackConfig);
        });
```

arches_her/settings.py
```diff
@@ -197,18 +197,27 @@
...
     from .settings_local import *
-except ImportError:
-    pass
+except ImportError as e:
+    try: 
+        from settings_local import *
+    except ImportError as e:
+        pass
 
 if DOCKER:
```
chrabyrd, line 213: what would happen here if there's no settings_local file with a DOCKER k/v ? Will it return falsy or throw a keyerror?

chrabyrd (changes requested):
overall looks good! Just a small change I have context about && a question

arches_her/settings.py
aarongundel: It's fine if there's no settings_local, but it won't work in docker.  settings_local is required for it to load the settings_docker file with the db connection info.

chrabyrd: ah the wrong line got highlighted -- this is about the `if DOCKER:` line. I was able to repro a break by doing an `if FOO:` line on my local

aarongundel: Well docker _should_ be set somewhere for aher.  In this case it's set in settings.py as False by default.

chrabyrd: oh I missed the `DOCKER = False` line above, this is good!

chrabyrd (approved):
lgtm!

chiatt (commented):
Looks good and works well for me.

chiatt (approved):

PR #1041, opened 2022-12-09
report bug fixes

johnatawnclementawn:
I found two reports where the assessmentCards were not being initialized properly and a few reports where scientificDateCards were initialized but never used. 

This also fixes an incorrect closing tag that was preventing some key-value data from being rendered properly.


johnatawnclementawn:
@aarongundel I can't assign this to you - I think I only have permissions for the AfS repo

aarongundel (approved):
looks good to me.

Issue #1044, opened 2023-02-02
Associated Consultations - More Accurate Message Based on Permissions


csmith-he:
In resource reports with associated consultation sections, if you don't have permissions to view consultations there will be no records.  The message in the section is confusing and if permissions are restricting access to this card, it may lead to the wrong conclusions. 

![image](https://user-images.githubusercontent.com/38533555/216313304-a9c6545f-4532-4854-9060-e2c8487f28ce.png)

If the user does not have permissions then the message should read "You do not have permission to view this data".

PR #1046, opened 2023-02-02
1044 - Change Default Message For Associated Consultations

csmith-he:
Remedies issue #1044

- resourceinstanceid object added to the resourceDataConfig for Activity, Area, Aircraft, Maritime, Monument and Person
- code added to provide a message to the user when they either do not have permissions to access associated consultation data or when there are no associated consultation records

### DIFF
--- a/arches_her/templates/views/components/reports/scenes/resources.htm
+++ b/arches_her/templates/views/components/reports/scenes/resources.htm
@@ -5,8 +5,13 @@
 {% block body %}
 <!-- Associated Activities section -->
 <div class="aher-report-section" data-bind="visible: !!dataConfig.activities">
-    <h2 class="aher-report-section-title"><i data-bind="click: function() {toggleVisibility(visible.activities)}, css: {'fa-angle-double-right': visible.activities(), 'fa-angle-double-up': !visible.activities()}" class="fa toggle"></i> {% trans "Associated Activities" %}</h2>
-    <a data-bind="{if: cards.activities, click: function(){addNewTile(cards.activities)}}"  class="aher-report-a"><i class="fa fa-mail-reply"></i> {% trans "Add Activity" %}</a>
+    <h2 class="aher-report-section-title">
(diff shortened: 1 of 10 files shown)


arches_her/templates/views/components/reports/scenes/resources.htm
```diff
@@ -343,11 +385,12 @@ <h2 class="aher-report-section-title"><i data-bind="click: function() {toggleVis
         <div class="aher-report-subsection">
             <div class="firstchild-container">
                 <div class="aher-table" data-bind="">
-                    <table class="table table-striped" cellspacing="0" width="100%">
+                    <p class="aher-table-aria-describedby" id="related-application-area-table_info">Related application area table.</p>
```
chiatt, line 388: I think this line is too close to the heading above it. Maybe it was in place for testing? It could probably just be removed. 

image

chiatt (changes requested):
This looks good and worked as expected for me, but I had to also add the `resourceinstanceid` to the `resourceDataConfig` of `consultation.js` as you've done for activity, area, historic-aircraft, etc.

csmith-he:
Hi @chiatt 
Thanks for getting back to me; I'll take a look at this tomorrow (19 Jul)

arches_her/templates/views/components/reports/scenes/resources.htm
csmith-he: @chiatt - Thanks for being patient while I took a look at this.

I've amended the consultation.js so the resourceinstanceid is in the resourceDataConfig.

I spoke to @phudson-he about that label.  It's there for accessibility reasons (screen readers) but should have been hidden by styling - I hadn't realised this as the styling was already present in our repo for Keystone but didn't seem to be in arches-her.  I've now added it to report_templates.css so the label should not longer be visible.

chiatt (approved):

Issue #1052, opened 2023-07-03
Related resource instance links always go to resource even in report view


csmith-he:
The link to a related resource instance always uses the /resource/ URL regardless of whether the user is in resource editor or report mode.  If a user only has read access to resources then this results in an error and being unable to navigate to the related resource.

The cause of this behaviour has been found in arches_her/media/js/utils/report.js from line 189.

The behaviour will be altered to check whether the URL of the current page is in resource editor or report mode and generate the correct related resource link accordingly.

PR #1053, opened 2023-07-31
1052 Build report or resource link based on originator URL

csmith-he:
Fulfils the requirements of #1052 

Modifies the getResourceLink function to check the urlPathname and use that to generate the resulting link with either report or resource in it.

This removes the issue of users being sent to the /resource/ URL even when they don't have edit access on the resource (which then results in an error).


chiatt (approved):
Looks good to me!

Issue #1056, opened 2023-09-12
BNG Datatype fails validation in Arches 6.1


chiatt:
The BNG Datatype fails to validate on save because it does not take a 'request' argument.

PR #1057, opened 2023-09-12
Prevent error raised during bngdatatype validation, re #1056

chiatt:
Prevent error raised during bngdatatype validation, re #1056


csmith-he (approved):
I'm happy this is validating properly now

PR #1069, opened 2024-03-04
Update index.htm for HER-demo

ekansa:
I will make an issue of adding a dismissible alert to the HER-demo home page.

A quick note however! I don't know where to put new image files so that they're in an expected directory location for the home page. Please feel free to copy this:

https://storage.googleapis.com/opencontext-media/her-demo/her-demo-map.png

and stick it into a reasonable and expected place.


aarongundel (approved):
LGTM

PR #1075, opened 2024-03-08
Her demo home page updates

ekansa:
I added some images to the (hopefully) appropriate media directory and added pointers to them using using the templating pattern like

```
{% static 'img/landing/her-demo-map.png' %}
```

Hopefully they work! If not, I can always go pack to pointing to the image files hosted on Google cloud storage (which is like s3)


aarongundel (approved):
lgtm!

PR #1115, opened 2024-04-12
#1102 Creates unique filename for letter using time as suffix

### LINKED ISSUE #1102: v75 Upgrade - Ensure all letters are created with unique name to handle browser caching behaviour
Change letter filename to include hour minute second and millisecond to handle browser caching behaviour (i.e. the user sees the latest letter generated, not the initial one cached by the browser).

Filename format going forward would be as follows:
< date > < letter_template_Name > < HMSmillisecond >.docx

csmith-he:
Fixes #1102


aj-he (approved):
works as expected.

PR #1118, opened 2024-04-15
#1109 Add deactivated spatial views

### LINKED ISSUE #1109: v75 Update - Add Deactivated Spatial Views
Include SQL in the AfHER package to load confirmed spatial views. These should be deactivated by default.

The solution will be to include the SQL in the post_sql directory of the package.

They should be loaded in a deactivated state so they don't affect data import performance.

csmith-he:
Fixes #1109 

Adds the generation of spatial views for the main resource models to a SQL file in the package, run as part of the post SQL process.

Sets the isactive value to false so they are deactivated by default.


aj-he (approved):
spatial views are loaded as expected

Issue #1104, opened 2024-04-11
v75 Upgrade - Fix broken workflows (widget data not loading)


csmith-he:
Workflows are broken - error in the console stating empty dictionary related to widgets is missing

Relates to: https://community.archesproject.org/t/issue-with-workflows-in-7-5/2270

aj-he:
The following code forces the `widgets` component to reload via webpack_require and then the workflows work ok.

This can be used to progress the rest of the upgrade work but we'll need to get to the bottom of why this is an issue.



![Image](https://github.com/archesproject/arches-her/assets/25295526/75b71003-3262-47cb-9a15-db0730c3e02a)

```js
if (Object.keys(widgets).length === 0) {
    console.error("[js/models/card-widget] widgets object is empty");
    if (__webpack_require__){
        try {
            widgets = __webpack_require__('../../arches/arches/app/media/js/widgets.js');
        } catch (error) {
            console.error(error);
        }
    }
}
```

aj-he:
In discussion with @chrabyrd a workaround has been added to fix an issue with widgets not loading correctly into the workflows. This will mean that we need to keep this under review as the workflow work progresses.

https://community.archesproject.org/t/issue-with-workflows-in-7-5/2270

aj-he:
fixed in #1122.

individual workflow fixes to follow

PR #1122, opened 2024-04-17
Initial fixes for Workflows

aj-he:
To enable the more complicated workflows to be fixed separately, this PR covers some general fixes as follows

- Adds fix for all workflows where the widgets were not being correctly loaded.
- Adds a workflow util that can be used to handle i18n strings if null.
- Adds workflow reversing functionality
- Fixes eslint errors.

### DIFF
--- a/arches_her/media/js/models/card-widget.js
+++ b/arches_her/media/js/models/card-widget.js
@@ -1,0 +1,196 @@
+define([
+    'underscore',
+    'knockout',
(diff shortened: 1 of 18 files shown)


csmith-he (approved):
I'm happy that this code is working up to the point we agreed for the initial fixes

csmith-he:
Initial fixes relating to #1104.  Further work required to complete issue fix.

PR #1139, opened 2024-04-19
#1134 Resource Descriptors removed

### LINKED ISSUE #1134: v75 update - display descriptors are broken
The custom display descriptor is returning [object Object] due to the intl string change

csmith-he:
Fixes #1134 

Handles display name, display description and map popup values in the new string format.


arches_her/views/resource.py
```diff
@@ -57,15 +57,24 @@ def get(self, request, resourceid=None):
...
+                map_popup = ''
                 if document['_source']['graph_id'] == '8d41e49e-a250-11e9-9eab-00224800b26d':
                     tiles = build_resource_dict([resource], active_cons_node_list, datatype_factory)
                     additional_data = tiles[0]
+                if document['_source']['displaydescription'][0]['value']:
+                    display_description = document['_source']['displaydescription'][0]['value']
+                if document['_source']['displayname'][0]['value']:
+                    display_name = document['_source']['displayname'][0]['value']
+                if document['_source']['map_popup'][0]['value']:
+                    map_popup = document['_source']['map_popup'][0]['value']
```
aj-he: this gets the first display descriptor. It should get the one for the active language
image

aj-he (changes requested):
use the active language to determine the value

aj-he:
@csmith-he I realised that this whole ResourceDescriptor was missing important content as the main RD has added userid and permissions info since first created. So I thought we could use the main RD as a base class, call the base get function and then inject the active_consultation data if it was a consultation.

I mocked up an example, but while testing I realised that the active consultations page was not using this. So I looked back at the blame and there used to be another search page for consultations that used a custom popup.

This functionality was removed in 2021, but the descriptor was not. Therefore, we could probably remove the whole custom descriptor. Can you test by commenting out the route in urls.py and seeing if there are any issues? If not, we should remove it.

csmith-he:
@aj-he No adverse effects I could see by removing the resource descriptor entirely and references to it so I've done that.

aj-he (approved):
It all works for me

PR #1152, opened 2024-04-26
#1150 Missing href="#"

### LINKED ISSUE #1150: v75 upgrade - Reports: add onEnterKeyClick and onSpaceClick for buttons without accessibility click
Follows on from #1140

It was found that some buttons in reports had no accessibility click ability so these need to be added

csmith-he:
Added missing href="#" to a tags in scene and report templates to allow them to be activated on an enter press.

Fixes #1150


phudson-he (approved):
LGTM

Issue #1101, opened 2024-04-11
v75 Upgrade - Create a hotfix for the spatialview database functions to support internationalisation and changes to resource descriptors


csmith-he:
- Basic support for internationalisation of the string datatype. This only support 'en' or the language of the first published graph found if english is not supported. Full support for multiple languages will require 7.6.

- Supports the retrieval of the display descriptor from the resource record rather than imitating the default display descriptor. This also will only support 'en' or the first other language found in the datatype.

NOTE: without the fix, string values appear as json string and the resource instance value is completely standalone and independent of anything set in the application. It is also susceptible to the json string value being in the resource-instance value

csmith-he:
Fixed by #1154

PR #1154, opened 2024-04-29
Adds hotfix for string and resource instance labelling in spatial views due to internationalisation

aj-he:
Spatial views functionality does not currently support the internationalisation of strings and the resource instance descriptors.

This replaces the three database functions that create labels for string, resource-instance and resource-instance-list.

It only supports one language as specified in the lang parameter default value, which is 'en'.

For Arches core 7.6, a lang code will be added to the spatial views definition, which will be passed to the functions.

Before this is approved, please ensure that it has been tested following a full package load from scratch, e.g. `python3 manage.py packages -o load_package -s /web_root/arches_her/arches_her/pkg -db -y`


csmith-he (approved):
Spatial views generating correctly as far as I can see.  LGTM

csmith-he:
Fixes #1101

Issue #1161, opened 2024-04-29
Source number in "Citation" report table


samuelscandrett:
Should the Source Number get added to the citations table in the report, so it can be visible and ordered by number - similar to a bibliography?  Currently the default sort order is alphabetical on the source name.

Example:
![image](https://github.com/archesproject/arches-her/assets/119508494/ebaa0ccd-d14e-4a3d-a3ee-5133bc5a0562)

Philbox:
This makes sense to me.

aj-he:
@SDScandrettKint - I guess the question is whether you need to sort by source then number, which that table won't allow. I have no opinion as to whether the column is visible or not.

aj-he:
fixed in #1165

PR #1165, opened 2024-04-30
Move Source Number into a column in the Citations table

samuelscandrett:
Addresses #1161 

Adds Source number to the Citations table and moves it out of the expanded data:
326588129-ebaa0ccd-d14e-4a3d-a3ee-5133bc5a0562


aj-he (approved):
Works as expected

PR #1172, opened 2024-05-01
#1167 Associated Construction Phase not visible in reports

### LINKED ISSUE #1167: Associated Construction Phase in Component doesn't show anything in report
Component with Associated Construction Phase node populated should show an association to that Construction Phase node in the report, but this appears as an empty field.

How it looks in the report:
image

Showing that there is an association in the resource editor:
image

csmith-he:
Fixes #1167 

Added associated monument construction phase to the testPaths for constuctonPhase in classifications.js


samuelscandrett (commented):
Works as expected

PR #1177, opened 2024-05-02
#1173 Hotfix for handling string in multi card step

### LINKED ISSUE #1173: v75 upgrade - multi-card step - edit button in cards causes error
appears to be similar the consultation date step which resulted in a card node that is a string not being made into an observable when being restored.  As that was a custom workflow step component we could check and make sure it got converted, however this is a core step so not clear how we can fix this?

Image

Image

csmith-he:
Fixes #1173 

Hotfix applied via a copy of workflow-component-abstract.js, specifically in the this.loadTileIntoEditor function.

Can now handle data in the new string format.

### DIFF
--- a/arches_her/media/js/views/components/workflows/workflow-component-abstract.js
+++ b/arches_her/media/js/views/components/workflows/workflow-component-abstract.js
@@ -1,0 +1,884 @@
+define([
+    'underscore',
+    'jquery',
+    'knockout',
+    'knockout-mapping',
+    'utils/workflows',
+    'arches',
+    'models/graph',
+    'viewmodels/card',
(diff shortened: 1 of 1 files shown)


aj-he (commented):
@csmith-he  This works for me.

Before I approve and merge, can you also put a comment at line 16 just above the `function NonTileBasedComponent()` to **highlight** that this file is included due to the hotfix at line xyz. I know there is already a comment down where the hotfix has been applied, It just means that anyone looking at the file for the first time will be able to see why it is here and it describes where the hotfix has been applied in the file.

I'm just thinking about what would help me if I where to look at this.

aj-he (approved):
LGTM

Issue #1135, opened 2024-04-19
v75 update Selected Consultation does not persist in Communication workflow once selected


csmith-he:
When a related consultation is selected in the first step of the Communication workflow, the selection does not persist when the following parts of the step are rendered in the browser:

Consultation selected:

![Image](https://github.com/archesproject/arches-her/assets/38533555/3a2068de-6605-4302-ab40-40eee9dd7e69)

Consultation selection not displayed when Date, Subject and Type are rendered visible as there is now a resourceid value in the view model:

![Image](https://github.com/archesproject/arches-her/assets/38533555/8d149622-2522-4fa4-a2fd-ea93c818bf92)

csmith-he:
@chiatt - Added you as an assignee for now after the chat over on Slack, basically so you can find it easily.

aj-he:
closed complete in #1178

PR #1178, opened 2024-05-03
Communication Workflow - resource select step viewmodel and step data fixes 

aj-he:
- Fixes the drop down losing the selected consultation after the resource is selected. (#1135)
- Fixes issues where the resource select step does not reload when moving back to it from notes step.
- Fixed i18n string for Digital Object name

### DIFF
--- a/arches_her/media/js/utils/workflows.js
+++ b/arches_her/media/js/utils/workflows.js
@@ -4,19 +4,24 @@ define([
 ], function(arches, ko) {
     return {
         getProp: (params, key, prop, isString=false) => {
+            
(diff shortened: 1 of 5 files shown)


csmith-he (commented):
@aj-he 

Looks like the issue in the screenshot below is back:

image

In templates/views/components/workflows/communication-select-resource.htm, the following change has been overwritten

image

(image from the PR that fixed this issue)

csmith-he (commented):
Fixes confirmed

csmith-he:
Fixes #1135

PR #1183, opened 2024-05-07
modifies status label from True/False to Active/Inactive

### LINKED ISSUE #1182: Consultation Status show true/false rather than Active/Inactive in report templates
Image

aj-he:
Changes the Consultation Status Label in report templates from True/False to Active/Inactive

fixes #1182 

image


csmith-he (approved):
Works as expected.  Approved.

PR #1185, opened 2024-05-08
Consultations and reports to use rem font sizes

### LINKED ISSUE #1147: v75 upgrade - check consultation and report-template fixed font sizes
font-size uses px in the consultations and report-templates - should use rem/em

phudson-he:
Change font sizes from px to rem for consultations and reports. This will ensure that font sizing scales as expected.

Fixes #1147


aj-he (approved):
Text appears to resize as expected in reports, details summary and the report on resource editor page

PR #1192, opened 2024-05-10
#1155 Refine available basemaps

### LINKED ISSUE #1155: v75 upgrade - refine available basemap layers
There are a lot of basemap layers which probably need to be removed or not loaded by default. Perhaps deactivate, zip or completely remove?

csmith-he:
Fixes #1155


aj-he (approved):
LGTM - if anyone needs the others then they can be re-added

Issue #1186, opened 2024-05-09
Consultation editor - card template in node tree not rendering correctly


aj-he:
When editing a consultation location node, the card templates in the tree fail to render...

![Image](https://github.com/archesproject/arches-her/assets/25295526/da1af2a7-cf0c-49c3-b313-642d88aa42df)

It appears to do with the related-resources-map card template - but it is not an issue when looking at it in the graph designer. It could be that it is only an issue with data?

aj-he:
The card tree renders correctly in the graph designer, but not in the resource editor. Looks like the cards might be in the wrong order in the resource editor and breaking in the image above, the graph designer has the Consultation Location card, but the editor is being passed the Current Basemap card

aj-he:
Issue traced to a commit 2 years ago. Changing the `self.form.cardComponentLookup[self.card.model.component_id()].componentname` on line 68 back to `self.form.cardComponentLookup[card.model.component_id()].componentname` (removing the `self.` when getting the `component_id()` ) stops the issue from happening.



![Image](https://github.com/archesproject/arches-her/assets/25295526/ae018ee4-e7fd-4e9a-8298-780af37622ba)

PR #1203, opened 2024-05-15
Fixes editor tree error when trying to load child cards for the consultation location top level card

aj-he:
Provides a fix for the issue in core where the Consultation Location card breaks the rendering of child cards in the editor tree #1186 .

The problem is that as the children where loaded, they were being passed the parent component name. This was an issue where they were different (related-resource-map card with default card children.

This will need to be fixed in core.


csmith-he (approved):
Checked against all available models to ensure it's working

PR #1206, opened 2024-05-20
#1196 Upgrade and add Generate Unique References function

csmith-he:
Generate unique references function added to project and package, and updated to handle new internationalisation format for strings.

Fulfils #1196

### DIFF
--- a/arches_her/media/js/views/components/functions/generate-unique-references-function.js
+++ b/arches_her/media/js/views/components/functions/generate-unique-references-function.js
@@ -1,0 +1,82 @@
+define(['jquery',
+    'knockout',
+    'knockout-mapping',
+    'views/list',
+    'viewmodels/function',
+    'templates/views/components/functions/generate-unique-references-function.htm',
+    'bindings/chosen'],
+function($, ko, koMapping, ListView, FunctionViewModel, generateUniqueReferencesTemplate, chosen) {
+    return ko.components.register('views/components/functions/generate-unique-references-function', {
+        viewModel: function(params) {
+            try{
+                FunctionViewModel.apply(this, arguments);
+                var self = this;
+                this.nodesSemantic = ko.observableArray();
+                this.uniqueresource_nodegroup = params.config.uniqueresource_nodegroup;
+                this.triggering_nodegroups = params.config.triggering_nodegroups;
+                this.simpleuid_node = params.config.simpleuid_node;
+                this.resourceid_node = params.config.resourceid_node;
+                this.nodesList = [];
+                this.nodesReference = params.config.nodegroup_nodes;
+
+                this.uniqueresource_nodegroup.subscribe(function(urng){
+                    self.nodesReference.removeAll();
+                    self.nodegroupid = urng;
+
+                    var filter = self.nodegroupid;
+                    
+                    if(!filter){
+                        return null;
+                    } else{
+                        ko.utils.arrayFilter(self.nodesList,function(item){
+                            if (item.nodegroup_id === filter){
+                                self.nodesReference.push(item);
+                            }
+                        });
+                    }
+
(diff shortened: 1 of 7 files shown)


csmith-he:
@SDScandrettKint would you be ok reviewing this PR?

arches_her/media/js/views/components/functions/generate-unique-references-function.js
```diff
@@ -0,0 +1,82 @@
...
+                            this.nodesSemantic.push(node);
+                        }  
+                    } else {
+                        this.nodesList.push(node);
+                    }
+                }, this);
+
+
+                this.graph.nodegroups.forEach(function(nodegroup) {
+                    if (nodegroup.nodegroupid in self.triggering_nodegroups){
```
samuelscandrett: I have noticed that the function regains a dirty state once saved, navigated away and then back to:

https://github.com/archesproject/arches-her/assets/119508494/d2c0fd07-2164-4a21-ad1c-2ec618b35b02

It appears that this foreach function is causing the `self.triggering_nodegroups` array to double in size upon the re-save of the tile, because the check of if the nodegroup is in the array isn't working.

Changing the following line ensures the list is not doubled in size and doesn't prompt the dirty state
```suggestion
                    if (self.triggering_nodegroups().includes(nodegroup.nodegroupid)){
```

samuelscandrett (changes requested):
This is really good!  From my testing it works as expected on resource creation through the UI and through CLI import (JSON and CSV).

Just a minor fix needed on the UI.

Should we also limit this function to the references nodegroup, similar to how the BNG functions only allow for those nodes (+ autofilled)?  Will it work if the incorrect nodegroup is selected?

Might also be nice to eventually add some documentation on how this function can be used for the release, but that can be planned later.

samuelscandrett:
During this review I noticed that the dropdown options are cut off towards the bottom @phudson-he 

This seems to happen more severely when the nodegroup has a larger number of nodes.  

1)
image

2)
https://github.com/archesproject/arches-her/assets/119508494/4592e445-4b7a-48c9-b0de-b397e5a7d014

3)
https://github.com/archesproject/arches-her/assets/119508494/b50967d2-e9d4-4bae-9447-da76c408d700

Further, there is a descrepency as to whether the dropdown has a search bar or not (comparing 2 and 3).  I can't determine if this is related to the number of nodes in a nodegroup (e.g. a nodegroup with <10 nodes doesn't have a search bar), however, I don't think it is the fault of this PR and instead a more general Arches issue.


EDIT:  It seems that there are two y axis scroll bars that influence this, example below:

https://github.com/archesproject/arches-her/assets/119508494/9c1b8d46-8472-4b8a-be70-8a95d9c7a791

csmith-he:
@SDScandrettKint 

I've added the fix you identified to stop the doubling of the triggering_nodegroups array - thanks for finding that and suggesting a solution.

I've also put a hotfix in the project.css to sort out the dropdown list being cut off so it works as expected for now.  I'll add an issue to the Roadmap Backlog for the dropdown to use SelectWoo rather than Chosen; doing that work now will probably hold us up with getting v1 of Arches for HERs out so we'll come back to that.

@aj-he and I had a chat about narrowing down the options in the dropdown lists.  For the moment, we'd like to keep things as is as it will allow people to use the function in other configurations if they end up making their own models - if it causes widespread problems we can always change this at a later date.

Would you mind reviewing again with the changes in, please?

samuelscandrett (commented):
All looks good to me!

One thing I notice is the node dropdown sometimes has a search bar and other times doesn't - however I think this might be a core feature (I'm guessing if nodegroup has > x nodes then show search bar), e.g.
image
image

CSS additions also work nicely.

samuelscandrett (approved):

PR #1212, opened 2024-05-28
Long search tags not wrapping

### LINKED ISSUE #1164: Search tag doesnt word wrap
Steps to re-produce:

1. Open search page
2. In the search box, find and select an item that is quite long

Results in the text not wrapping...

Image

phudson-he:
Upon further investigation, it's not going to be wise to cause wrapping of the search tags as it breaks the containing elements. selectWoo is used throughout the site and it's not worth causing a huge amount of work checking for various situations its being used in etc. Best option was to display ellipses indicating there is more text, and this is the selectWoo default method.

Fixes #1164

### DIFF
--- a/arches_her/media/css/project.css
+++ b/arches_her/media/css/project.css
@@ -1007,6 +1007,8 @@ span.rp-tile-title {
     padding: 0px;
 }
 
+/* Ensures search tags gracefully display any text overflowing */
+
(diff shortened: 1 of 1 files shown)


csmith-he (approved):
Appears to be working as expected.  Happy to approve

image

Issue #1214, opened 2024-05-30
Add Letter template cleaning tool as management command


aj-he:
The docx files used for letters can have the field tags broken by the style runs when the file_template view processes them. This is because the style runs can split the tag name, and stop the find and replace process from identifying them.

HE has a script to fix this, but it should be added as a management command to allow this to be done more easily.

This should be added to the readme.md in the repo so that users know that they need to run this whenever they update the letters.

aj-he:
Being worked on by @RobOatesHistoricEngland

PR #1224, opened 2024-06-05
Add management command to fix Word letter template style runs

RobOatesHistoricEngland:
#1214 

Preserving style runs in the letter templates allows for richly formatted letters to be produced. However, a downside to this approach is that the template's fields, which get replaced by data, cannot have any changes in the styling. Sometimes, even when the styling has been corrected, Word can still produce multiple runs of text.

Due to the fragmentation of the field tag (e.g., `<Contact Name>`), it can be internally referenced in the DOCX file as `<Contact` and `Name>`. This means that when `file_template.py` is performing the search and replace operation, `<Contact Name>` is not found, resulting in `<Contact Name>` remaining literally in the generated correspondence letter.

This helper management command will parse all the .DOCX files in a given directory (default is `docx`) and join back together any field tags that have been split apart due to styling changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,4 +73,16 @@ Administrators of an instance of Arches for HERs can configure their implementat
 
 
 
+## Working with Letter Templates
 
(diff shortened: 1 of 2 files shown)


csmith-he (approved):
Checked this helper command works as expected using the directions in the readme

PR #1234, opened 2024-06-25
Active consultations - apply correct date sorting and check for None

### LINKED ISSUE #1209: Check Active Consultation sorting is not correct
It was noticed that sorting didn't appear to be correct for months and days, but years did.

Example from GLHER...
 
Steps to reproduce:
Navigate to active cons
Search active cons for 'Adam'
Sort by Log date newest to oldest
Click into each result to see log date.
The order of the dates I saw for first 10 were:
2022-06-09
2022-06-09
2022-06-10
2022-06-10
2022-06-10
2022-08-03
2021-04-29
2020-10-16
2014-08-13
2014-07-16

RobOatesHistoricEngland:
#1209

The lambda expression was not handling `None` values correctly and was also failing to convert string representations of dates into a sortable format, regardless of the international date formats used in different installations.

### DIFF
--- a/arches_her/views/active_consultations.py
+++ b/arches_her/views/active_consultations.py
@@ -25,14 +25,13 @@ from arches.app.models.resource import Resource
 from arches.app.models.tile import Tile
 from arches.app.datatypes.datatypes import DataTypeFactory
 import json
-import re
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


csmith-he (approved):
Looks to be working as expected

PR #1235, opened 2024-06-25
Add skip to content to index page

### LINKED ISSUE #1240: Skip to content needed on index page for accessibility
Adds a "skip to content" link on index for navigating using keyboard and screen reader

phudson-he:
To fix an accessibility testing issue raised re no bypass block on the homepage, these changes will add the "skip to content" feature, upon first tab press. This will bypass the top menu items.

It's worth noting that I've had to use the trans tags for the text, as the $root.translations method isn't available on the index page.

fixes #1240


aj-he (approved):
skip to content works for me, and is described over voiceover

PR #1241, opened 2024-07-02
Changes size of caption blocks and updates alignment

chiatt:
Changes the caption block size so that the caption text is on only two lines, not three.


aarongundel (approved):
LGTM

PR #1242, opened 2024-07-02
1173 - Remove workflow component hotfix

csmith-he:
Hotfix now added to Core in https://github.com/archesproject/arches/pull/11103 so no longer required in Arches for HERs


aj-he (approved):
Multi card step editing in workflows work with hotfix removed

PR #1269, opened 2024-07-12
#1264 Aria Label and Empty Link Accessibility Errors

### LINKED ISSUE #1264: Accessibility- Report Page- Activity: 1.3.1- Info and relationships- Empty links and broken Aria Labels
Across several pages there are empty links that sit at the end of some headings as well as several broken Aria references.

The empty links are at the end of these headers:
- Names and Identifiers> External cross References
- Names and Identifiers> System Reference Numbers
- Location Data> Geospatial Coordinates> Geometry
- Archive Holding> Activity Archive Material
- Associated People and Organizations> Associated People and Organizations
The broken Aria references can be found on all tables from what I have seen, I will attach screenshots of where WAVE has pointed to these issues

1.3.1- Info and Relationships
Logical structure involves checking that the screen reader reads out what a user with good vision would be able to see, and in the right order (E.G format changes and and structure of the page are visual, headings etc)

csmith-he:
Fixes #1264 

- Set info in defaultTableConfig to true (set to false causes aria-describedby error).  This does mean "showing x to x of x entries" is visible beneath tables but perhaps this can be hidden visibly by styling in another PR (@phudson-he FYI)
- Creates a new function, checkCardsAvailable, in report.js which can be used to see if there are cards in the data or parent object; if there are then the user is in resource manager mode and if not then the user is in report mode
- Only displays the add/edit links if there are cards in the data or parent object, as checked using the checkCardsAvailable function (this stops the links from rendering in the code if the user is in the report view, thus fixing the Empty Link Accessibility errors)

### DIFF
--- a/arches_her/media/js/utils/report.js
+++ b/arches_her/media/js/utils/report.js
@@ -113,7 +113,7 @@ define([
             paging: false,
             searching: false,
             scrollCollapse: true,
-            info: false,
+            info: true,
             columnDefs: [{
                 orderable: false,
                 targets: -1,
@@ -147,6 +147,13 @@ define([
                 if(tile){
                     tile.selected(true);
                 }
+            }
+        },
+
+        // Can be used to check if current page is in report mode or resource manager mode
+        checkCardsAvailable: function(cards){
+            if(Object.keys(cards).length > 0){
+                return true;
             }
         },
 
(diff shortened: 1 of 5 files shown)


phudson-he (approved):
LGTM

Issue #1273, opened 2024-07-15
Artefact type Confidence/Certainty node missing


samuelscandrett:
Missing an Artefact Type "Confidence" or "Certainty" node. 

This node could be similar to "Phase Certainty" in the same nodegroup:

![image](https://github.com/user-attachments/assets/66520162-7edc-4d62-ad27-67bbf5433570)

PR #1281, opened 2024-07-22
Add Artefact Type Certainty node

samuelscandrett:
Addresses [#1273](https://github.com/archesproject/arches-her/issues/1273)

Adds Artefact Type Certainty and Artefact Type Certainty Metatype nodes on the same level as Artefact Type.
image
image

Resulting card looks as follows:
image

@Philbox Need I'm sunsure whether Phase Classification is the correct parent nodegroup for Artefact Type Certainty, or if you think Artefact Type would be a better parent for it?
image

### DIFF
--- a/arches_her/media/js/views/components/reports/scenes/classifications.js
+++ b/arches_her/media/js/views/components/reports/scenes/classifications.js
@@ -31,7 +31,7 @@ function(_, ko, arches, reportUtils, ClassificationsTemplate) {
 
             self.artefactProdTableConfiguration= {
                 ...self.defaultTableConfig,
-                columns: Array(13).fill(null)
+                columns: Array(14).fill(null)
             };
             // Components Table
             self.componentsTableConfig = {
@@ -171,6 +171,7 @@ function(_, ko, arches, reportUtils, ClassificationsTemplate) {
                         const phaseDescription = self.getNodeValue(node, 'phase classification', 'phase classification description', 'phase description');
                         const phaseEvidence = self.getNodeValue(node, 'phase classification', 'phase evidence type');
                         const startDate = self.getNodeValue(node, 'production time span', 'from date');
+                        const artefactTypeConfidence = self.getNodeValue(node, 'phase classification', 'artefact type certainty')
                         const tileid = self.getTileId(node);
 
                         return {
@@ -189,6 +190,7 @@ function(_, ko, arches, reportUtils, ClassificationsTemplate) {
                             phaseEvidence,
                             productionTechnique,
                             startDate,
+                            artefactTypeConfidence,
                             tileid
                         };
                     }))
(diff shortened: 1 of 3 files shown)


csmith-he:
@SDScandrettKint Working as expected in the UI.  Just to check, this Artefact Type Certainty isn't expected to be displayed in the reports, is it?

samuelscandrett:
> @SDScandrettKint Working as expected in the UI. Just to check, this Artefact Type Certainty isn't expected to be displayed in the reports, is it?

Thanks for spotting this @csmith-he, I forgot about adding it to the report!  It has been added in the latest commit ([855507a](https://github.com/archesproject/arches-her/pull/1281/commits/855507a52934a0a461a12c5b5bd3c8a6d52c50e4)).  I don't think it's necessary for it to have its own column, so I have added it in the "show details" part.

csmith-he:
Just to confirm from a functionality standpoint, I've managed to load our test cut of GLHER data into an instance of AfHERs with these changes in the models without any ill effects that I can see.

From this point on, it just needs to be looked at by @Philbox now

Philbox:
@SDScandrettKint Hang this off artefact type rather than production. Otherwise it's fine.

aj-he:
@SDScandrettKint - Once you have changed the parent node of the new Artefact Type Certainty node as suggested by @Philbox, we can get this merged.

samuelscandrett:
As of [681a0f2](https://github.com/archesproject/arches-her/pull/1281/commits/681a0f20f87f61ae750063596dd3a710caf4fe5b), Artefact Type Certainty (+ metatype) have been moved under Artefact Type:
image

Both Artefact Type Certainty and Artefact Type Certainty Metatype have the Ontology Class E55_Type and the relation to parent (Artefact Type) P2_has_type.  Previously, Artefact Type Certainty had the relation of P141_assigned, but it is no longer available in the dropdown and hardcoding it in the model JSON causes errors 
image

aj-he:
@SDScandrettKint - I've asked @Philbox to take a look

Philbox:
@SDScandrettKint the P141 assigned property is no longer true because that relates to the classification assignment. P2 has type is the correct property.

So E55 Artefact Type -> P2 has type -> E55 Artefact Certainty Type

This way the certainty type has a semantic relationship with Artefact Type and not any other node in the Phase Classification.

aj-he:
Review on hold until graph changes discussed and agreed by group

chiatt:
It seems like this should include a migration to apply any graph changes to existing instances.

chiatt (changes requested):
A migration to add these graph changes is necessary to update existing systems.

samuelscandrett:
> A migration to add these graph changes is necessary to update existing systems.

@chiatt Definitely agree. Would each graph change need it's own migration file, or could we group all of the approved changes into one migration?

chiatt:
> > A migration to add these graph changes is necessary to update existing systems.
> 
> @chiatt Definitely agree. Would each graph change need it's own migration file, or could we group all of the approved changes into one migration?

I think one migration file would be fine.

PR #1283, opened 2024-07-22
Add Advice Start Date and End Date as columns to Consultation Advice table

### LINKED ISSUE #1282: Not possible to sort Advice by start and end dates in Consultation report
Advice start and end dates are viewable in the resource view but not the report.  
image

It would be useful to show these two values in the report to allow the advice stages to be sorted by date.

samuelscandrett:
Closes #1282

Adds columns for Advice start and end dates like so:
image

These columns allow the different Advice tiles to be sorted by either start or end date.


csmith-he (approved):
LGTM

PR #1285, opened 2024-07-26
Fix for report tabs being read out using screen reader

### LINKED ISSUE #1267: Accessibility- All resource reports: 4.1.2- Name, Role, Value- Selected tab not announced by screen reader
**When** I navigate over tabs
**Then** I am only informed that the tabs are clickable and not which tab I currently have selected which is only visually signified by being underlined

Screen readers should be able to tell that a tab is currently selected or it leaves the user unsure of which tab they are currently on

**4.1.2- Name, Role, Value**

phudson-he:
Fixes issue with screen reader when navigating report tabs, affects all reports. Added required aria-controls/expanded values to identify what is being displayed and aria-live to force opened content to be read out. Tested with NVDA, now reads as expected.

Fixes #1267


csmith-he (approved):
LGTM

PR #1294, opened 2024-08-08
Update index page with responsive version

chiatt:
- Updates the index page with the responsive version created by @ekansa
- Re-adds the welcome popup and makes it responsive.


aarongundel (approved):
LGTM

PR #1296, opened 2024-08-08
Fix length of "Resource Referenced By" table 

### LINKED ISSUE #1277: Adding Associated Consultation in Application Area resource causes console error
Adding an associated consultation in an Application Area causes report to break
image

samuelscandrett:
Addresses #1277

The "Resource Referenced By" table didn't have a fixed length because the rows were added asynchronously, which was causing issues with paging and thus the other resource association tables.

Changing `<!-- ko if: getRelatedResources() -->` to `<!-- ko if: relations().length -->` seems to give the table a correct length, allowing the others to render properly.

### DIFF
--- a/arches_her/media/js/views/components/reports/scenes/referenced-by.js
+++ b/arches_her/media/js/views/components/reports/scenes/referenced-by.js
@@ -70,6 +70,8 @@ define([
                         // error
                     });
             };
+
+            self.getRelatedResources();
         },
         template: ReferencedByTemplate
     });
--- a/arches_her/templates/views/components/reports/scenes/referenced-by.htm
+++ b/arches_her/templates/views/components/reports/scenes/referenced-by.htm
@@ -12,7 +12,7 @@
     <div class="aher-nodata-note">{% trans "No resource referenced by for this resource" %}</div>
     <!-- /ko -->
 
-    <!-- ko if: getRelatedResources() -->
+    <!-- ko if: relations().length -->
     <div class="aher-report-subsection">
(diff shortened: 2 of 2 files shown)


csmith-he (commented):
@SDScandrettKint - is there a reason why line 11 isn't following the same pattern as line 15?

aj-he:
@csmith-he @SDScandrettKint - the use of `getRelatedResources()` twice in the template, plus @SDScandrettKint's comment that the length of the `relations()` observableArray makes me think that the function should be called at the initialisation of the viewmodel, and then `relations` used in by the ko if conditions?

csmith-he:
Further note, if there is an application area that isn't associated with any consultations, the getRelatedReferences function runs successfully and there is no length to relations() which results in this:

image

csmith-he:
@SDScandrettKint as per @aj-he's suggestion, could you add the following to the viewmodel and the template to see if it all behaves properly for you (it's behaving as I'd expect with and without consultations that have the app area associated with them):

referenced-by.js
image

referenced-by.htm

image

samuelscandrett:
Works nicely! @csmith-he @aj-he 

Length is correctly set to resources referenced by, and original error no longer occurs.  When no resources are referenced, the table correctly shows the "No resources referenced..." message.

Code changes have been added in [71a360d](https://github.com/archesproject/arches-her/pull/1296/commits/71a360d7526e6927caf697fb2a2c6ba4fb25eb6d).

csmith-he (approved):

Issue #1292, opened 2024-08-07
Resource delete button produces console error 


samuelscandrett:
The delete button in the resource table produces a console error once pressed, and the UI only reflects the row deletion once the page has been refreshed, and the tile is still visible in the node tree:

https://github.com/user-attachments/assets/a71d8aac-78b4-4ca8-8c23-0fbb9aa15865

This was not the case in the version 6 instance of AfHER.  In this version the row would be visibly deleted, and the tile would also disappear from the node tree:

https://github.com/user-attachments/assets/35faa1bc-4325-45d9-980f-cd3e1a17ee10

aj-he:
Completed by #1337

PR #1298, opened 2024-08-09
Fix for ajax console error when deleting tile/row from table

samuelscandrett:
Closes: #1292 

The change in jquery versions between Arches version 6 and 7 means that the `.success` method no longer works; now success is a part of the ajax call.

The deletion ajax call now works, and the deletion is reflected in the table and node tree.

### DIFF
--- a/arches_her/media/js/utils/report.js
+++ b/arches_her/media/js/utils/report.js
@@ -40,12 +40,13 @@ define([
             return $.ajax({
                 type: "DELETE",
                 url: arches.urls.tile,
-                data: JSON.stringify(tile.getData())
-            }).success(() => {
-                const tiles = card.tiles();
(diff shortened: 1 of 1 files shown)


samuelscandrett:
This change has uncovered a new bug when moving away and back to the Associated resources tab, where the table length is not updated, causing UI and console errors as shown below:

https://github.com/user-attachments/assets/ae7d2f02-739d-4d73-8b09-48d3d6981148

This issue is not new to the v7 AfHER branch and was present in v6.

Note that the error is the same as issue #1277, confirming that this error refers to the table length

aj-he:
@SDScandrettKint - this was raised but then you pointed out addition issues. Is there anything else to come?

aj-he:
Closed in favour of PR #1337 which includes additional fixes for the delete button

PR #1301, opened 2024-08-13
Ensure collapsable report sections announce state changed by screen reader

### LINKED ISSUE #1289: Accessibility - Report section collapse expand button not read out to screen readers
**When** I navigate through the headings on the resource reports
**Then** The screen reader announces the expand/collapse box only as "clickable" or "expand/collapse" rather than if it is expanded or collapsed

This could leave the user unsure if the heading is accessible of they have accidentally collapsed it, meaning they are unable to access its content

**4.1.2 – Name, Role, Value**

aj-he:
Adds an aria attribute handler binding and is applied to all collapsible sections across reports and scenes

fixes #1289


phudson-he (dismissed):
LGTM - so far - but the clickable i element needs a... aria-controls="relevant_unique_ID" and the relevant div that gets displayed needs a corresponding unique id

aj-he:
Passed internal testing

phudson-he:
Happy to merge without the aria-control values/corresponding id's, seeing as its passed testing

PR #1303, opened 2024-08-15
Further fix for screen reader announcement that tab displayed

### LINKED ISSUE #1267: Accessibility- All resource reports: 4.1.2- Name, Role, Value- Selected tab not announced by screen reader

phudson-he:
This fix ensures that when a report tab is selected, that the screen reader announces that a new tab is currently being displayed.

Fixes #1267

### DIFF
--- a/arches_her/templates/views/components/reports/scenes/archive.htm
+++ b/arches_her/templates/views/components/reports/scenes/archive.htm
@@ -6,7 +6,9 @@
 
 <!-- Bibliographic Source Creation section -->
 <div class="aher-report-section" data-bind="visible: !!dataConfig.sourceCreation">
-    <p class="sr-only" data-bind="text: document.querySelector('.aher-report-anchors li.active').innerText + ' tab now displayed'"></p>
+    <p class="sr-only" data-bind="let: {tabTitleText: document.querySelector('.aher-report-anchors li.active').innerText + ' tab now displayed'}">
+        <span data-bind="text: {% trans "tabTitleText" %}"></span>
+    </p>
     <h2 class="aher-report-section-title"><i tabindex="0" data-bind="onEnterkeyClick, onSpaceClick, click: function() {toggleVisibility(visible.sourceCreation)}, css: {'fa-angle-double-right': visible.sourceCreation(), 'fa-angle-double-up': !visible.sourceCreation()}"  class="fa toggle"></i> {% trans "Source Creation" %}</h2>
     <!-- ko if: checkCardsAvailable($data.cards) -->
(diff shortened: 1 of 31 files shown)


arches_her/templates/views/components/reports/scenes/archive.htm
```diff
@@ -6,6 +6,7 @@
 
 <!-- Bibliographic Source Creation section -->
 <div class="aher-report-section" data-bind="visible: !!dataConfig.sourceCreation">
+    <p class="sr-only" data-bind="text: document.querySelector('.aher-report-anchors li.active').innerText + ' tab now displayed'"></p>
```
aj-he: @phudson-he  - can you wrap "tab now displayed" in a `{% trans  "" %}` tag

something like...
```
<p class="sr-only" data-bind="text: document.querySelector('.aher-report-anchors li.active').innerText + '{% trans " tab now displayed" %}'"></p>
```

aj-he:
@phudson-he - also, if multiple scenes are used in the same tab, what happens? Should this just be added as the first element of the tabcontent... eg in consultations...

image

arches_her/templates/views/components/reports/scenes/archive.htm
phudson-he: Added the trans tag in commit https://github.com/archesproject/arches-her/pull/1303/commits/b23b9f074d7590189d0c9956dcc1f5dccbad7153

phudson-he:
@aj-he Good spot for the consultation report, now fixed in commit https://github.com/archesproject/arches-her/pull/1303/commits/b5849bcc45e66c86eced46ed927f227567b13a20 - some of the consultation report details are already catered for in the scenes, but some sections had to be fixed accordingly

aj-he:
@phudson-he I only used consultations report as an example. There may be others like that? I didn't think the scenes needed the screen reader elements as the tabs and containers are in the reports files so they should go there?

aj-he (approved):
@phudson-he - This works for me and the expanded/collapsed states read out correctly for me. However, you pointed out that some of the tab content fails to read out when expanded. This I could recreate using VoiceOver on Mac but only **after the first time** the tab content was made active.

This seems to only occur when there is a table at the top of the first section, and only on some tabs.

Therefore, I think the issue is related to the table and outside the scope of this ticket.

PR #1304, opened 2024-08-16
Updated demo home page to fix grey area placements

ekansa:
The update is visible for testing here: https://arches.opencontext.org/

I've been checking it on BrowserStack and it seems to work OK. However, I think we shouldn't try to force the text areas over the image to have a consistent height, since it tends to break responsiveness.


chiatt (approved):
LGTM!

PR #1308, opened 2024-08-19
Updates invalid default node values with valid values used in other graph files

### LINKED ISSUE #1307: Location Data branch file has missing default values for some nodes
Running the following SQL shows that there are some node default values that don't exist

```sql
SELECT valueid, value, conceptid, default_values.*
    FROM values v 
	RIGHT OUTER JOIN (
		SELECT g.name as graph, g.isresource, n.name as node, cnw.nodeid, cnw.config ->> 'defaultValue' AS default_value
		FROM cards_x_nodes_x_widgets cnw 
		JOIN nodes n on n.nodeid = cnw.nodeid
		JOIN graphs g on n.graphid = g.graphid
		WHERE n.datatype in ('concept', 'concept-list')
		AND cnw.config ->> 'defaultValue' != ''
	) default_values 
	ON v.valueid::uuid = default_values.default_value::uuid
	--WHERE v.valueid IS null;
```

Image

aj-he:
Updates default values of some nodes in the in the Location Data branch file (not where used in resource models) with valid values. These are taken from the `pkg/graphs/branches/Geometry.json` and `pkg/graphs/branches/Area.json` files which are valid.

image

fixes #1307

### DIFF
--- a/arches_her/pkg/graphs/branches/Location Data.json
+++ b/arches_her/pkg/graphs/branches/Location Data.json
@@ -557,7 +557,7 @@
                 {
                     "card_id": "b1d2cd15-f446-11eb-8b1c-a87eeabdefba",
                     "config": {
-                        "defaultValue": "b022dbc0-b722-4e12-a675-12fca24ddffa",
+                        "defaultValue": "1d5c7465-7225-4bca-b1df-045a86e286b2",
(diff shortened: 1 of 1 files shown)


chiatt (approved):
Looks good!

PR #1311, opened 2024-08-21
Ensures bng buffer info readout on input focus

### LINKED ISSUE #1310: BNG Filter - buffer info not read out
Ensure that on entering the bng buffer input, it reads out the info panel as with the grid ref and XY inputs

aj-he:
fixes #1310 and reads out input info panel for buffer.


phudson-he (approved):
Good to go!

Issue #1319, opened 2024-08-28
Consultation Outcomes nodegroup lacks notes node


samuelscandrett:
The Consultation model's Outcomes nodegroup is missing an "Outcome Notes" node for recording reasons behind decisions and outcomes.

**Note**: Since this is a graph change, it requires input/approval from the wider community before progression can be made.

samuelscandrett:
As discussed with @Philbox on 20/12/24, rather than just adding an Outcome Notes nodegroup off Outcomes like in #1320, the PR for this issue should extend the Outcomes nodegroup to document a log history of outcomes.  

We should use the Timespan (EDTF) branch (http://127.0.0.1:8000/graph_designer/7541f73e-839c-11ea-b880-f875a44e0e11) and add a rich-text string node for notes.

PR #1320, opened 2024-08-28
Add Outcome Notes nodes to Consultation model

samuelscandrett:
Addresses #1319

Adds the following nodes to the Outcome nodegroup:
image
I used the Audit metadata notes nodes and the Artefact discovery notes as a reference for the convention of adding notes nodes.

The card added is just the rich-text note:
image


samuelscandrett:
Since this is a graph change, it requires input/approval from the wider community before progression can be made.  This PR has been reverted to a draft and moved into the "Requires graph changes" section of the project.

PR #1325, opened 2024-09-04
Fix to hide map buttons and display action buttons

### LINKED ISSUE #1321: Can't save tiles when creating a new Application Area through the Consultation workflow
Creating a new Application Area through the consultation workflow & map card appears to be broken, as tiles can't be saved, and sections of the map card still persist:

image

When adding new Application Areas through the regular Consultations resource editor, sections of the map card still persist, but data is able to be saved:

image

Note, this is not the case for other "Create new X resource" in the workflows e.g. creating a new Maritime Vessel in the Application Area workflow "Related Heritage Resources" stage:

image

phudson-he:
This change ensures correct z-index for the map filter buttons so that they aren't displayed over the top of the Create New Application Area panel, when creating a Consultation Workflow. Also to ensure that the previously hidden action buttons are not hid on this particular panel.

**Note: This needs to be tested by a tester before merging, which I will arrange once approved**

Fixes #1321

### DIFF
--- a/arches_her/media/css/project.css
+++ b/arches_her/media/css/project.css
@@ -824,7 +824,7 @@ table.report-tabular-card {
     margin-top: 10px;
 }
 
-.tabbed-workflow-step-body .workbench-card-sidepanel .install-buttons {
+.tabbed-workflow-step-body .workbench-card-sidepanel .install-buttons:not(.tabbed-workflow-step-body .workbench-card-sidepanel .create-resource-instance-card-component .install-buttons) {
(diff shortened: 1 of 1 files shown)


csmith-he (approved):
LGTM

samuelscandrett (approved):
Works nicely!  It allows Application Area resources to be created whilst in the workflow, and the edit tab from the map editor no longer leaks through.

phudson-he:
Tests passed by internal tester!

PR #1331, opened 2024-09-06
1326 Add Spinny When Waiting for Correspondence 

### LINKED ISSUE #1326: Correspondence workflow briefly returns to first step when generating letter
When continuing to the next step after selecting a consultation resource and letter type, the workflow returns to the first step with each of the nodes empty for a brief moment.  This is the case until the letter has completed generating, then the final step is loaded.  This can be slow (a few seconds), potentially depending on if the consultation resource or the letter template is complex.  

During this slow stage, a user may become confused and think that the workflow has failed and attempt to re-enter the data.

An example video is shown below:

https://github.com/user-attachments/assets/7e50c3e4-9ed2-4d75-8b0e-328a7e6b525c

The duration of the letter generation and thus the first page reappearing can also depend on the network speeds.  See the following example video using throttling on Firefox where it is shown for longer than the first video.

https://github.com/user-attachments/assets/c81642b8-0dd7-4c32-979f-ab947af51533

This links to a similar issue where if a consultation + letter type is selected but fails to generate the letter, the workflow will return to the first step without an error message.

csmith-he:
Closes #1326 

When user presses the save button in the Correspondence Workflow a spinny with the message 'Generating correspondence for consultation' appears.

This clears once the correspondence is generated or the process fails.

### DIFF
--- a/arches_her/media/js/views/components/workflows/correspondence-select-resource.js
+++ b/arches_her/media/js/views/components/workflows/correspondence-select-resource.js
@@ -74,7 +74,7 @@ define([
                         'resourceXresourceId': ''
                     }];
 
-                    params.pageVm.loading('Generating correspondence for consultation...');
+                    
 
                     $.ajax({ //saving the realted resource (digital object) to the Letter node (consultation)
                         url: arches.urls.api_node_value,
@@ -92,7 +92,6 @@ define([
                             console.log("Digital related resource updated");
                         })
                         .fail(function(response) {
-                            params.pageVm.loading(false);
                             console.log("Updating digital related resource failed: \n", response);
                         });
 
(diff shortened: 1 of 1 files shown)


samuelscandrett:
The PR works well, however, I've noticed that when throttling the network through Firefox, the "Select Related Consultation" dropdown reappears for a short period of time still.   It reappears between the "Save and continue button" being clicked and [line 77](https://github.com/HistoricEngland/arches-her/blob/0513652050a9add027ce126d583c1a2d093acda9/arches_her/media/js/views/components/workflows/correspondence-select-resource.js#L77) where the message and loading are fired, which is much shorter than before but can still be confusing for the user.  See video for an example:

https://github.com/user-attachments/assets/91db855c-3cbe-4640-8dd6-c97857fb02e7

<br />
                     
However, if the line `self.saving(true);` is added after `self.tile().save().then(function(tile) {` as shown in the screenshot below...



... then the "Select Related Consultation" dropdown is hidden much quicker (almost a flash):

https://github.com/user-attachments/assets/79801b04-f73b-4cfc-90f4-3995bacfb358

samuelscandrett (approved):
Looks good to me!

aj-he (approved):
Looks ok to me. @SDScandrettKint can I leave you to check and complete the PR?

samuelscandrett (approved):
Works nicely, happy to merge.

Issue #1332, opened 2024-09-09
AfHER using LatLong/WGS84 ANALYSIS_COORDINATE_SYSTEM_SRID rather than BNG


samuelscandrett:
The settings.py variable ANALYSIS_COORDINATE_SYSTEM_SRID is set to lat/long rather than BNG.  This means that any map filtering either via the BNG search feature or the default map filter is done to Lat/Long and different to BNG.

aj-he:
completed by #1333

PR #1333, opened 2024-09-09
Set the analysis coordinate sys SRID to BNG

samuelscandrett:
Addresses #1332


aj-he (approved):
I think that given the Preferred Coord System default is BNG and the BNG filter, which is installed out of the box, requires it then this should be the default value.

PR #1335, opened 2024-09-11
1317 MVT Application Areas with permissions

### LINKED ISSUE #1317: Map_sources using /geojson endpoint causing performance issues

csmith-he:
fixes #1317 
Uses code provided by @chiatt in #1318 but works with a project setup
Also includes check that user has permissions to see the map source


csmith-he:
@SDScandrettKint and @chiatt 

Putting together this PR so that it will work with the project setup @SDScandrettKint current has in action.  Sam, can you give this a whirl and see if it meets your needs so you don't have to use pg_featureserv?

@chiatt, the idea would be to have the migration and the things currently in the select_sources.sql in your PR in the Arches Application way of doing things.  Does that sound do-able?

If you're both happy, I'll change this from a draft to an active PR.

arches_her/pkg/graphs/resource_models/Consultation.json
```diff
@@ -52,7 +52,7 @@
                         "overviewzoom": 11,
                         "pointradius": 4,
                         "selectRelatedSource": "select-application-area",
-                        "selectRelatedSourceLayer": "",
+                        "selectRelatedSourceLayer": "app-area",
```
chiatt, line 55: This should probably be done with the migration rather than changing the graph directly, otherwise upgrading instances will be difficult.

arches_her/pkg/preliminary_sql/select_sources.sql
```diff
@@ -1,7 +1,9 @@
 INSERT INTO map_sources(name, source)
```
chiatt, line 1: I think this is okay for now, but it will need to be moved to a migration as HER is transitioned from a project to an Arches Application.

arches_her/views/map.py
```diff
@@ -0,0 +1,41 @@
...
+class ApplicationAreas(View):
+    def get(self, request, zoom, x, y):
+        nodeid = '1909956f-3a3b-11eb-ae99-f875a44e0e11'
+        if hasattr(request.user, "userprofile") is not True:
+            models.UserProfile.objects.create(user=request.user)
+        viewable_nodegroups = request.user.userprofile.viewable_nodegroups
+        
+        try:
+            node = models.Node.objects.get(nodeid=nodeid, nodegroup_id__in=viewable_nodegroups)
+            resource_ids = get_restricted_instances(request.user, allresources=True)
```
chiatt: You probably don't want to use 'allresources' here. That will filter out any resource that any user does not have access to. It's really a problem with the api.py view that was put into place to support caching because of performance issues with that view. However, this view is much faster and caching isn't needed, so you can filter out just those instances that the current user is not allowed to see. You can see how I did this in my PR.

chiatt (changes requested):

arches_her/pkg/preliminary_sql/select_sources.sql
chiatt: @csmith-he Ah.. just saw your comment above. You're already on it.

arches_her/pkg/graphs/resource_models/Consultation.json
csmith-he: Understood - I think @SDScandrettKint wasn't sure if the migration would work in the project instance he currently has ESCC's Arches for HERs instance running on

arches_her/pkg/preliminary_sql/select_sources.sql
csmith-he: @chiatt Definitely the plan for it to be moved to a migration for the Arches Application

arches_her/views/map.py
csmith-he: @chiatt Thanks for the explanation - that really helps.  I'll make the changes as needed.

arches_her/pkg/graphs/resource_models/Consultation.json
csmith-he: @chiatt - Is this still something that needs to be applied by migration?

aj-he: @chiatt @csmith-he - as this is part of a graph JSON in the package for the 1.0.0 release, it doesn't need a migration. All future graph JSON changes after release 1.0.0 will need to be accompanied by a migration, which the arches app pattern will allow us to do.

arches_her/views/map.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class ApplicationAreas(View):
+    def get(self, request, zoom, x, y):
+        se = SearchEngineFactory().create()
+        restricted_resource_ids = get_restricted_instances(request.user, search_engine=se)
+        if len(restricted_resource_ids) == 0:
+            restricted_resource_ids.append("10000000-0000-0000-0000-000000000001")  # This must have a uuid that will never be a resource id.
+        restricted_resource_ids = tuple(restricted_resource_ids)
+        nodeid = '1909956f-3a3b-11eb-ae99-f875a44e0e11'
```
aj-he: @csmith-he where did the node permission check go? This only checks resource instance instance permissions

arches_her/pkg/preliminary_sql/select_sources.sql
```diff
@@ -1,7 +1,9 @@
...
     VALUES ('select-application-area', '{
-        "data": "/geojson?nodeid=1909956f-3a3b-11eb-ae99-f875a44e0e11&include_geojson_link=true",
-        "type": "geojson"
+        "data": "/geojson?nodeid=1909956f-3a3b-11eb-ae99-f875a44e0e11", 
+        "type": "vector", 
+        "tiles": ["/application-areas/{z}/{x}/{y}.pbf"], 
+        "minzoom": 6
     }');
 
 INSERT INTO map_sources(name, source)
```
aj-he: `select-area` and `select-monument` are no longer needed and can be removed from the SQL script

aj-he:
@csmith-he - apart from the `select-area` and `select-monument` sources needing to be removed, this works for me.

aj-he (approved):
LGTM

aj-he:
@chiatt @SDScandrettKint - You are both in the reviewer list for this so I won't merge until you've reviewed or removed yourself as a reviewer

aj-he:
chiatt agreed on a call to proceed so merging.

arches_her/pkg/preliminary_sql/select_sources.sql
aj-he: sql file is now finalised and will be upgraded to migration separately

arches_her/views/map.py
aj-he: suggested change has been applied

Issue #1334, opened 2024-09-10
Delete button in behaviour only suited to cardinality n


aj-he:
Discovered while fixing #1292 

In the editor, DataTables in the reports have a delete button on rows to delete the tile for that row.

The data tables are used on two ways
1. to represent a tile per row on an n cardinality card
2. to represent an item in a list datatype node (`concept-list`, `resource-instance-list`) in a card with cardinality 1.

The delete button uses the tileid associated with the row and deletes the tile.

Delete works fine for rows representing 'n' cardinality cards, but for a '1' cardinality card where rows represent items in a list node, clicking delete deletes the tile and therefore all rows are deleted.

Need to remove the Delete (trash) button from tables for cardinality != 'n' for now to protect against accidental deletion and then look at how to handle list nodes

aj-he:
Fixed in #1337

PR #1337, opened 2024-09-12
Fix for report DataTable delete button issues

### LINKED ISSUE #1292: Resource delete button produces console error

aj-he:
Fixes #1292 #1334 

- Delete button uses new jquery ajax pattern (@SDScandrettKint)
- Deleting a tile using this button now updates the report resource object so that the datatable can render correctly when moving between tabs.
- The delete button has been removed from tables representing list datatype nodes (row per item in the node list) as deleting 1 row actually deletes all data in the table. This is because it deletes the tile containing the list rather than the item.  In future a new button can be created that handles this scenario but for no it is removed to stop unintended data loss.

There is a known issue with the nested tables in Consultation Site Visits. The delete button only works on those rows if the page is refreshed after the Site Visit subcards are created. This will be raised as a separate issue.


aj-he:
@SDScandrettKint - hopefully this make things more stable. I don't get breaking tables anymore.

@csmith-he - could you please sanity check this too.

samuelscandrett (approved):
Looks good to me.  Testing tile deletion works, and changing back and forth tabs no longer shows the error of incorrect table lengths.

csmith-he (approved):
LGTM

PR #1342, opened 2024-09-18
1340 Remove tab stops from letter templates

### LINKED ISSUE #1340: Letter templates have an erroneous tab stop
The docx templates for the correspondance workflows have a very large tab stop in them, which causes letters to indent too much if a tab character is included in the text (such as consultation advice).

example WSI Amend template

(template)
Image

(letter output)
Image

csmith-he:
Tab stops removed from letter templates to fix the way the templates handle injected text that has a tab within it.

fixes #1340


aj-he (approved):
I have visually checked the templates and can see no tab stops. Generating the letters now have a sensible indent.

The indented (numbered) lines wrap to the left margin instead of the indent, but I think that is a separate enhancement that can be looked at later.

PR #1356, opened 2025-01-03
Fixes for 76 search_filters and create_spatial_views sql files

### LINKED ISSUE #1355: 76_upgrade branch - issues with preliminary_sql/search_filters.sql and post_sql/create_spatial_views.sql
Errors in the SQL files when loading up the AfHER app package on the 76_upgrade branch.

## 1. preliminary_sql/search_filters.sql 
```
loading preliminary sql
  /home/samuel/arches_dev/arches_her/AfHER_application/arches_her/arches_her/pkg/preliminary_sql/search_filters.sql
column "sortorder" of relation "search_component" does not exist
LINE 10:                 sortorder,
                         ^

Failed to load sql files
```

Looks like `sortorder` is not a column in the search_component db table anymore, same with `enabled`, as of [this PR](https://github.com/archesproject/arches/pull/10807) and [these lines](https://github.com/archesproject/arches/blob/4dec2b6c6df9e9fe3d1aecec338bb0a02cdbbb54/arches/app/models/migrations/10804_core_search_filters.py#L96-L103).  Similarly, config is now a column in the table.

Solution would be to remove these lines from the search_filters.sql insert, and add an empty config object.

## 2. post_sql/create_spatial_views.sql 
```
loading post sql
  /home/samuel/arches_dev/arches_her/AfHER_application/arches_her/arches_her/pkg/post_sql/create_spatial_views.sql
null value in column "languageid" of relation "spatial_views" violates not-null constraint
DETAIL:  Failing row contains (289d89bb-47ad-40ef-89fe-3df03c265135, public, maritime, Used to record the details of historic vessels which are either ..., f, [{"nodeid": "656c68c4-3ec0-11eb-859a-f875a44e0e11", "description..., f, 9f07fa25-f457-11eb-98c7-a87eeabdefba, null).

Failed to load sql files
```

This looks to be caused by the missing `languageid` column ([added here](https://github.com/archesproject/arches/commit/41e253c06fa78412e72a9fe2b110e407edb33631)).  Solution would be to add "en" as a languagecode to all sp_views.

samuelscandrett:
Closes #1355 and #1358

Removes sortorder and enabled, and adds an empty config object to preliminary_sql/search_filters.sql

Adds languageid and 'en' to all sp_view rows of post_sql/create_spatial_views.sql

Should the config for BNG filter be populated?


samuelscandrett:
I am still getting issues with BNG filter:
```
loading preliminary sql
  /home/samuel/arches_dev/arches_her/AfHER_application/arches_her/arches_her/pkg/preliminary_sql/search_filters.sql
duplicate key value violates unique constraint "search_component_componentpath_key"
DETAIL:  Key (componentpath)=(views/components/search/bng-filter) already exists.

Failed to load sql files
```

Is it populating from 0001_initial.py?  Is search_filters.py required anymore?  
The `SearchComponent.objects.update_or_create` in 0001_initial still includes sortorder and enabled, so do we need a new migration to update this (and potentially others)?

<br>
Lastly, post_sql only runs successfully if the spatial_views_intl_hotfix.sql file is deleted, @aj-he are we able to delete this hotfix file for 7.6?

aj-he:
@SDScandrettKint - given the new pattern of using migrations in apps, I suspect that the inclusion of sql files in the package should be removed in favour of having them being in a migration file, or as files used by a migration process as per @bferguso proposal here: https://github.com/archesproject/arches/issues/11696.

With reference to your specific question on the spatial view hotfix, this can be removed from 7.6 onwards. However, can you check that an existing 7.5 instance with hotfix in place can be migrated to this new version. If there is an issue, you may need to add a reversible migration to handle any issues.

We need to make sure we include release instructions if users need to make any local changes. @chiatt could you advise what needs doing for this?

We also need to bump the arches_her version number. There is a question as to how we bump it? Given Arches moving up a minor version, we should do the same? Again, @chiatt, any thoughts on this?

samuelscandrett:
@aj-he latest commit [559c981](https://github.com/archesproject/arches-her/pull/1356/commits/559c9813514589b31e23359920238b5ae8f607c9) adds the start of a 0002 migration (may need a better name).  Currently this migration just adds the search overlay map layers.  Previously these map layers were populated with a random UUID, so I've done a name search incase the layers already exist (as we don't want duplicates), but if not then assigned them a set UUID.  I'm unsure if we want it to continue using a random uuid1 or if a set UUID is better?  
I've also removed the spatial view hotfix, but still need to test upgrading from 7.5 -> 7.6.

All preliminary_sql files can be included in a migration, but the post_sql files are a little more tricky, as they rely on the graphs/nodes already populated (spatial views require geometry nodes, and updating the graphs table with the report templates requires a populated graphs table).  Should these remain in post_sql, or is there a better way we can manage this within a migration?

chiatt:
> @SDScandrettKint - given the new pattern of using migrations in apps, I suspect that the inclusion of sql files in the package should be removed in favour of having them being in a migration file, or as files used by a migration process as per @bferguso proposal here: [archesproject/arches#11696](https://github.com/archesproject/arches/issues/11696).
> 
> With reference to your specific question on the spatial view hotfix, this can be removed from 7.6 onwards. However, can you check that an existing 7.5 instance with hotfix in place can be migrated to this new version. If there is an issue, you may need to add a reversible migration to handle any issues.
> 
> We need to make sure we include release instructions if users need to make any local changes. @chiatt could you advise what needs doing for this?
> 
> We also need to bump the arches_her version number. There is a question as to how we bump it? Given Arches moving up a minor version, we should do the same? Again, @chiatt, any thoughts on this?

Yeah - I would bump this up a minor version given the significance of the changes. Regarding documenting the release - for AfS we added a [releases directory](https://github.com/archesproject/arches-for-science/blob/dev/2.0.x/releases/2.0.1.md), following the same pattern as core Arches, so I think that would be a good pattern for AfHERs as well.

chiatt (changes requested):
Rather than querying map layers based on name, it would be more dependable to query based on the source property found in the layer definition json (`search-results-points`, `search-results-hashes`, `search-results-hex`).

samuelscandrett:
> Rather than querying map layers based on name, it would be more dependable to query based on the source property found in the layer definition json (`search-results-points`, `search-results-hashes`, `search-results-hex`).

In [67c315b](https://github.com/archesproject/arches-her/pull/1356/commits/67c315b78d964e5c95dc3ca3bac2f0fb4e02a7b6) I replaced searching for the name with searching the layerdefinition source attribute.  
This has been tested with setting up the db via the package, migrating with the map_layers already in, and migrating with the map layers having changed their name.

I also added a new migration file (which could possibly be squashed into 0002 within this PR) for adding the BNG filter into the new search component structure.  This migration needs to come after the core migration `10804_core_search_filters`, however, django currently isn't playing nice and isn't allowing this to work as it should when setting up the db from scratch with the package.  Instead, it pushes all arches_her migrations forward to 10804, which causes an error with 0001.  
If AfHER has already been installed up to 0001_initial, then these migrations work as intended.

arches_her/migrations/0003_add_bng_search_view.py
```diff
@@ -0,0 +1,73 @@
+from django.db import migrations, models
+from django.utils.translation import gettext as _
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("arches_her", "0002_move_pkg_sql"),
+        ("models", "10804_core_search_filters"),
```
chiatt: I think this `10804_core_search_filters ` dependency may cause problems because it is already a dependency of another arches migration, `11117_add_bulk_concept_editor`. 

I think it should probably be changed to the last migration in arches 7.6 (`11499_add_editlog_resourceinstance_idx`) and moved to be the first dependency of `0002_move_pkg_sql.py`. That way you finish Arches 7.6 migrations, and then resume the arches_her migrations at `0002_move_pkg_sql `.

In other words, make sure Arches updates to `11499_add_editlog_resourceinstance_idx` before `0002_move_pkg_sql` runs, then run `0003_add_bng_search_view`.

samuelscandrett:
Latest commit sorts out the migration flow problems I have been facing.  Using run_before (as well as the core Arches dependency) we can control where the migration occurs, and sort of fence it in.  Thankfully, this means we don't have to create conditional logic for 0001, or make any real changes to the code.  @csmith-he just copying you in here so you're aware of the progress, as you were present in the stand-up where we spoke about this issue.

Using `python manage.py showmigrations --plan` we can see the flow is correct, with Arches 7.5 completing its migrations before arches_her 0001 runs.  Then, core 7.6 runs before finishing off with arches_her 1.1.x migrations 0002 and 0003.

```
[X]  models.9945_file_thumbnail_bin_file_thumbnail_text
[X]  arches_her.0001_initial
[ ]  models.9946_alter_notification_context
[ ]  models.10515_i18n_plugins
[ ]  models.9191_string_nonlocalized
[ ]  models.10709_refresh_geos_by_transaction
[ ]  models.10710_fix_whatisthis
[ ]  models.10798_jsonld_importer
[ ]  models.10999_add_principaluserid_to_resources
[ ]  models.10999_update_principaluser
[ ]  models.10799_geojsongeometry_featureid
[ ]  models.10516_rename_geometrynodeid_spatialview_geometrynode_et_al
[ ]  models.10705_spatialviews_internationalisation_support
[ ]  models.11179_file_and_geom_search
[ ]  models.10804_core_search_filters
[ ]  models.11117_add_bulk_concept_editor
[ ]  models.10476_add_user_bulk_data_import
[ ]  models.11499_add_editlog_resourceinstance_idx
[ ]  arches_her.0002_move_pkg_sql
[ ]  arches_her.0003_add_bng_search_view
```

I think this PR might now be reaching the end.
@aj-he I've tested that spatial views created in 7.5 + AfHER v1.0 work in 7.6 + AfHER v1.1.  Are there more specific scenarios we need to test?  
Are we fine with there being two new migration files, or should 0003 be merged into 0002?

aj-he (changes requested):
@SDScandrettKint - the CI errors seem to be occuring due to the package.json commands have not been correctly set on lines 13-14. Could you fix?

https://github.com/k-int/arches-her/blob/407c39838041b69705bb22b13d06af765d3e8bd7/package.json#L13

aj-he (approved):
OK, I am happy that there is an action to fix up the tests in a separate issue/PR as it'll pollute this PR with black formatting fixes if we do it now.

I'm also happy with the submitted solution so approving.

@chiatt - you're down to review too, are you ok with @SDScandrettKint fixing tests separately?

chiatt (commented):
This looks good. I'm wondering, however, if we should remove the reverse migration. Some of these removals, such as datatypes, won't work because they have relationships to the graph.

samuelscandrett:
I've removed the 0001 reversal migrations for widgets, datatypes and report templates as they are all required by the AfHER graphs. The other extensions: plugins, search components and functions have remained because they are not directly required by the graph out of the box. The BNG search component doesn't use the datatype/widget.

chiatt (changes requested):
This looks really good. Just one small change and I think it should be done. If you reverse migrate and then migrate forward again, the migration breaks because the bngcenterpoint already exists and the update throws an integrity error because we use a manually specified primary key which `update_or_create` does not like. 

To get around this you could use defaults:
```
        Datatype.objects.update_or_create(
            pk="bngcentrepoint",
            defaults={
                "iconclass":"fa fa-location-arrow",
                "modulename":"bngcentrepoint.py",
                "classname":"BNGCentreDataType",
                "defaultwidget":bngpoint,
                "defaultconfig":None,
                "configcomponent":"views/components/datatypes/bngcentrepoint",
                "configname":"bngcentrepoint-datatype-config",
                "isgeometric":False,
                "issearchable":True,
            },
        )
```


Or you could except the IntegrityError:

```
from django.db.utils import IntegrityError

        try:
            Datatype.objects.update_or_create(
                datatype="bngcentrepoint",
                iconclass="fa fa-location-arrow",
                modulename="bngcentrepoint.py",
                classname="BNGCentreDataType",
                defaultwidget=bngpoint,
                defaultconfig=None,
                configcomponent="views/components/datatypes/bngcentrepoint",
                configname="bngcentrepoint-datatype-config",
                isgeometric=False,
                issearchable=True,
            )
        except IntegrityError:
            pass
```

chiatt (approved):

chiatt:
> This looks really good. Just one small change and I think it should be done. If you reverse migrate and then migrate forward again, the migration breaks because the bngcenterpoint already exists and the update throws an integrity error because we use a manually specified primary key which `update_or_create` does not like.
> 
> To get around this you could use defaults:
> 
> ```
>         Datatype.objects.update_or_create(
>             pk="bngcentrepoint",
>             defaults={
>                 "iconclass":"fa fa-location-arrow",
>                 "modulename":"bngcentrepoint.py",
>                 "classname":"BNGCentreDataType",
>                 "defaultwidget":bngpoint,
>                 "defaultconfig":None,
>                 "configcomponent":"views/components/datatypes/bngcentrepoint",
>                 "configname":"bngcentrepoint-datatype-config",
>                 "isgeometric":False,
>                 "issearchable":True,
>             },
>         )
> ```
> 
> Or you could except the IntegrityError:
> 
> ```
> from django.db.utils import IntegrityError
> 
>         try:
>             Datatype.objects.update_or_create(
>                 datatype="bngcentrepoint",
>                 iconclass="fa fa-location-arrow",
>                 modulename="bngcentrepoint.py",
>                 classname="BNGCentreDataType",
>                 defaultwidget=bngpoint,
>                 defaultconfig=None,
>                 configcomponent="views/components/datatypes/bngcentrepoint",
>                 configname="bngcentrepoint-datatype-config",
>                 isgeometric=False,
>                 issearchable=True,
>             )
>         except IntegrityError:
>             pass
> ```

Error doesn't seem to occur for @samuelscandrett, so I think it's fine to disregard it as an environmental issue on my end.

PR #1366, opened 2025-09-08
Fix bng validation bug

### LINKED ISSUE #1350: BNG map filter doesn't accept coordinate with leading zeros
If an X or Y coordinate begins or ends with a zero, the search won't accept the number due to the error "X and Y must both be 6 digit coordinates", as shown below.  The 0 cannot be removed, as then it doesn't meet the 6 number criteria.

      

Oddly, if the grid reference that is created from one of these coordinates with a leading 0 is used, the search is successful but the same error of "X and Y must both be 6 digit coordinates" is displayed.  Example:

https://github.com/user-attachments/assets/c50f0f5c-005f-4438-a734-5855dad23afa

This means if you want to search a BNG that includes a leading 0, the grid reference must be used.

CWDamm-Kint:
Fixes bug https://github.com/archesproject/arches-her/issues/1350, where eastings and northings beginning with zero were not accepted. 

The `validateXy` method already checks that x and y (eastings and northings) are numeric, so it's okay to check the string length rather than casting to an integer first. This  then prevents the leading zero from being dropped. 

Note that it still shows the error until you've entered both x and y. 

https://github.com/user-attachments/assets/e0201d94-d6a5-4448-aac6-1e338f7608c5

Closes https://github.com/archesproject/arches-her/issues/1350,


chiatt (changes requested):
This looks really good. However, if a user enters a coordinate that begins with 0 and then refreshes the page, the zero is stripped from the coordinate on reload.

CWDamm-Kint:
I think this should keep the zeros after refreshing now. The x and y coords are reconstructed from the BNG reference in the url, so it was a case of converting them back to strings and adding the leading zeros with padding. 

Whilst working on it I came across https://github.com/archesproject/arches-her/issues/1403, so it might be worth merging  https://github.com/archesproject/arches-her/pull/1404 before finalising this one.

samuelscandrett:
This works and looks good to me.

I've noticed two additional bugs with the BNG search, but I'll raise them in separate issues.

PR #1378, opened 2025-09-24
fix BNG filter in 7.6

### LINKED ISSUE #1359: BNG Filter errors in 7.6
In v7.6 the BNG filter doesn't appear to be working as intended:

Image

This also prevents any search results from appearing.

samuelscandrett:
Closes #1359

### DIFF
--- a/arches_her/media/js/views/components/search/bng-filter.js
+++ b/arches_her/media/js/views/components/search/bng-filter.js
@@ -336,8 +336,7 @@ define([
                     return bng;
                 };
 
-
-                this.filters[componentName](this);
+                this.searchFilterVms[componentName](this);
                 this.pageLoaded = false;
                 var bufferColour = "#009ab9";
                 var gridSquareColour = "#f0c200";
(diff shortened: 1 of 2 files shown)


samuelscandrett:
[32b12f0](https://github.com/archesproject/arches-her/pull/1378/commits/32b12f003e120cb69d3fcf0bfcdba66489f2e614) fixes the initial /search error, so the search page now loads, but the map view in the BNG filter still doesn't work:

image

samuelscandrett:
This PR is now on pause until https://github.com/archesproject/arches-her/issues/1380 is resolved. This is the reason why the BNG filter map does not appear in the search component.

chiatt (approved):
lgtm

chiatt:
> This PR is now on pause until #1380 is resolved. This is the reason why the BNG filter map does not appear in the search component.

It seems like this should be merged ahead of resolving #1380 simply because it looks like these changes will need to be made anyway and it would be helpful to have search working.

samuelscandrett:
@chiatt this makes sense to me. https://github.com/archesproject/arches-her/issues/1380 should fix the map issue without any further input here, and if there is we can just open a new PR.

Issue #1376, opened 2025-09-23
Update AfHER 7.6 index.htm


samuelscandrett:
As stated in the description of https://github.com/archesproject/arches-her/pull/1345#issue-2557771325, the index and docker files need updating to work with 7.6

samuelscandrett:
Index.htm currently looks as so, with the Firefox error: 
`The stylesheet http://127.0.0.1:8000/static/css/project.scss was not loaded because its MIME type, "application/octet-stream", is not "text/css".`


<img width="1918" height="823" alt="Image" src="https://github.com/user-attachments/assets/d6a00486-3229-451c-a06d-25cb3fb98410" />

With accessibility mode switched on we also see a similar error 
`The stylesheet http://127.0.0.1:8000/static/css/accessibility.scss was not loaded because its MIME type, "application/octet-stream", is not "text/css".`

PR #1379, opened 2025-09-26
Fix for index page problems

CWDamm-Kint:
Scss files were not importing correctly: https://github.com/archesproject/arches-her/issues/1376
 
To fix this, Index.htm has been updated to use `render_bundle` to import the css files from webpack, rather than the scss files directly. This is inline with the index file for core Arches 7.6: https://github.com/archesproject/arches/blob/dev/7.6.x/arches/app/templates/index.htm

In addition, the font-size setting from index.css is interfering with the styling in Arches for HER's index page. This was previously countered in project.scss, but this approach is no longer possible, as project.scss also impacts other pages in Arches, such as the search page. 

After some investigating, we've determined that index.css isn't needed at all, so its import has been removed, preventing the problem from occurring to begin with.


phudson-he (approved):
LGTM

Issue #1380, opened 2025-09-26
project.css isn't working anymore


samuelscandrett:
The project.css file imported 
```
@import url(./report_templates.css);
@import url(./bng-filter.css);
@import url(./accessibility-plugin.css);
```

Now in 7.6, report templates and the bng filter aren't working, as shown below and in https://github.com/archesproject/arches-her/issues/1359, respectively.

<img width="1263" height="826" alt="Image" src="https://github.com/user-attachments/assets/2075fc4b-0863-477d-94b6-e8d28aa71f45" />
    

     
Moving everything into project.scss seems to work, is this the approach? If so, I think we could include it in https://github.com/archesproject/arches-her/pull/1379 which is already fixing some CSS files and related issues.

CWDamm-Kint:
I've had a look and I think I agree with moving the imports into project.scss. Project.css isn't being loaded at all anymore and can probably be deleted. 

Whilst we are at it though, I think it might be worth removing all of the landing-page css into a separate css file within the app (landing-page.css). This is only really needed within the arches_her app, and specifically only for the custom index.htm, so this should make it a bit clearer what's going on. 

Do you have any more thoughts on this @phudson-he?

phudson-he:
@CWDamm-Kint  I've done the same, as in made a separate landing page scss file. Here's the structure:

<img width="237" height="246" alt="Image" src="https://github.com/user-attachments/assets/991ae908-6c2d-4d29-a29d-1f90406c4f64" />

...so /media/css/project.scss is getting processed using `{% render_bundle 'css/project' 'css' %}`. The content of project.scss is just `@import "project-files";` and this in turn pulls in the project-files/_index.scss which contains...

<img width="203" height="86" alt="Image" src="https://github.com/user-attachments/assets/3288836d-60ec-4e7c-a300-e76651306691" />

As you can see there are a few separate scss files for general, login, landing and spinner. These are just tailored for us at HE but this is where project wide styling changes can be made, if required. Obviously all the custom HE landing page styling is in _landing.scss. For the landing page I made sure that styling only affects the landing page by adding a wrapper class to the body tag like this:

<img width="294" height="24" alt="Image" src="https://github.com/user-attachments/assets/56be55ba-64f1-4396-8e5d-ecb652f83607" />

..then everything in _landing.scss is inside the body wrapper class...

[...]

PR #1388, opened 2025-10-09
Add new css file structure

CWDamm-Kint:
This replicates the structure used on the Mariner project: https://github.com/HistoricEngland/mariner-proj/tree/aj_release/1.0.0_with_core_dev_7.6.x/mariner_proj/media/css

As discussed in https://github.com/archesproject/arches-her/issues/1380#issuecomment-3355869102, project.css has been removed, and `project.scss` now just imports the directory project-files. The `_index.scss` of this page then imports the partials, including `_landing.scss`. 

I've changed the following files to scss, as I was otherwise  struggling to import them into the index file:

- 'report_templates';
- 'bng-filter';
- 'accessibility-plugin';

Similarly, I'd have rather imported `project-functions` and `project-variables` within `_index.scss`, but this wasn't working in the way you would expect, even when imported before `_landing.scss`. I suspect it may be due to webpack trying to separately compile all the imported variables within `_landing.scss`, before everything has been imported into `_index.scss`. 

Closes: #1388


chiatt (approved):
This looks good to me!

phudson-he (approved):
LGTM!

samuelscandrett (approved):
This looks good! 
Also fixes issues with the BNG filter and report templates.

samuelscandrett:
I've noticed a side effect from this PR where navbar option menu headers are larger than expected. 

This seems to be coming from the following CSS property in project.css
```
.menu-title {
  font-size: 1.7rem;
}
```

image

image

CWDamm-Kint:
I think this is because consultations.scss is now being imported into _index, following this PR: https://github.com/archesproject/arches-her/pull/1392. I'm not sure it was actually being used before? 

.menu-title is setting the size of those headers to 1.7 rem. If that class isn't being used anywhere in the actual HER consultations, we could just remove it from consultations.scss, or change the font-size to 1.3rem?

samuelscandrett:
@CWDamm-Kint thanks, this makes sense, I must have checked-out the wrong commit and assumed it had come from this PR!
Please could you add your findings to #1395 ?

PR #1391, opened 2025-10-10
Add migration for including "show": true to Accessibility Statement and Active Consultations plugins

### LINKED ISSUE #1390: Active Consultations and Accessibility Statement plugins no longer showing in navbar
Both the Active Consultations and Accessibility Statement plugins are no longer showing in the navbar.

They are both registered as plugins in the db, as done in the first migration.

However, both are missing `"show": True,` from their config.

samuelscandrett:
Closes #1390 

Adds a 0004 migration file for adding "show" to the plugin config.

The one major issue still outstanding is that `plugin.config.pop("show", None)` and `plugin.save()` isn't actually saving the removed property from the config. Printing the config after removal does reveal that the show key has been removed, so I'm not sure why saving isn't working.

Attempting removal with `del` and `I18n_String.__delitem__(x.config, "show")` also didn't work

### DIFF
--- a/arches_her/migrations/0004_update_plugins.py
+++ b/arches_her/migrations/0004_update_plugins.py
@@ -1,0 +1,33 @@
+from django.db import migrations, models
+from django.utils.translation import gettext as _
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("arches_her", "0003_add_bng_search_view")]
+
+    def update_plugins(apps, scheme_editor):
+        Plugin = apps.get_model("models", "Plugin")
+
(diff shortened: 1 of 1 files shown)


samuelscandrett:
In [ef130e0](https://github.com/archesproject/arches-her/pull/1391/commits/ef130e09ba80d9cbf074c326dcb3de2478a5af2c) I changed the migration code from Python to SQL in order to successfully reverse the state of both plugin configs.

chiatt (approved):
lgtm!

aj-he (approved):
lgtm too!

PR #1392, opened 2025-10-10
fixes broken consultation/app area workflow

aarongundel:


chiatt (approved):
Looks good - fixes workflows

PR #1394, opened 2025-10-13
Changes the order of the declarations to avoid deprecation warnings

### LINKED ISSUE #1393: Landing page is causing webpack warnings
The landing page css is causing a lot of warnings in webpack, due to the ordering of the declarations. Should be a quick fix. 

Example:

Image

CWDamm-Kint:
Changes the order of the declarations in the landing page css, so that un-nested declarations come before nested ones. 

Closes https://github.com/archesproject/arches-her/issues/1393


samuelscandrett (approved):
LGTM

PR #1401, opened 2025-10-21
HER controlled lists migration

alexwuthrich:


arches_her/settings.py
```diff
@@ -168,6 +200,11 @@ def get_optional_env_variable(var_name, default=None) -> str:
     "django_hosts",
     # "silk",
     "arches_her",
+    "django.contrib.postgres",
+    "arches_querysets",
+    "arches_component_lab",
+    "arches_controlled_lists",
```
johnatawnclementawn, line 204: To be able to call `packages -o import_controlled_lists`, you'll need to move `arches_controlled_lists` above `arches`.

Here's an [example](https://github.com/archesproject/arches-lingo/blob/e17dee62616066ab7f90a63d51c1d79c4de6e351/arches_lingo/settings.py#L243-L270)

```suggestion
    "arches_her"
    "webpack_loader"
    ...
    "arches_controlled_lists",
    "arches"
    "arches.app.models",
    "arches.management",
    ...
```

chiatt:
Currently this PR does not migrate the nodes identified in #1402 to the reference datatype because they do not have valid collections assigned to them.

pyproject.toml
```diff
@@ -28,6 +28,7 @@ requires-python = ">=3.11"
 dependencies = [
   "arches>=8.0.4,<8.1.0",
   "python-docx==1.1.0",
+  "arches_controlled_lists @ git+https://github.com/archesproject/arches-controlled-lists.git@main",
```
johnatawnclementawn:
```suggestion
  "arches_controlled_lists @ git+https://github.com/archesproject/arches-controlled-lists.git@dev/1.0.x",
```
We removed the main branch in favor of development branches

chiatt (changes requested):
Now that the controlled list package loader supports skos files, I think it would good to re-export the lists as skos rather than .json so that they are available on package load.

PR #1404, opened 2025-11-04
Adds turf buffer units object

### LINKED ISSUE #1403: Bounding box is not being created for 12 character BNG references
When entering a 12 character BNG reference, `turf.buffer` is used to create a bounding box around the point that's identified. 

This is currently throwing an error:

https://github.com/user-attachments/assets/9b70b14b-987d-4312-bf5e-48b30330cb0d

I think this can be fixed with a relatively small change to how 'meters' is being provided as an argument. The turf docs suggest it needs to be wrapped in an object: https://turfjs.org/docs/api/buffer

CWDamm-Kint:
This brings turf.buffer syntax inline with the correct syntax: https://turfjs.org/docs/api/buffer

'meters' is now wrapped in an object: `{units: 'meters'}`.

Closes https://github.com/archesproject/arches-her/issues/1403

### DIFF
--- a/arches_her/media/js/views/components/search/bng-filter.js
+++ b/arches_her/media/js/views/components/search/bng-filter.js
@@ -546,7 +546,7 @@ define([
 
                 try {
                     self.map().getSource('grid-square')?.setData(geoJSON);
-                    bounds_geojson = isPoint ? turf.buffer(bounds_geojson, 100, 'meters') : bounds_geojson;
+                    bounds_geojson = isPoint ? turf.buffer(bounds_geojson, 100, {units: 'meters'}) : bounds_geojson;
                     var extent = geojsonExtent(bounds_geojson);
                     var bounds = new self.mapboxgl.LngLatBounds(extent);
                     self.map().fitBounds(bounds, {


samuelscandrett:
LGTM, this prevents the error.

@aj-he should issues like this one and #1403 target a 1.0.1 patch, or do we want to prioritise 1.1.0? I suppose this opens up the wider question of what becomes AfHER LTS which we could discuss in the dev call.

aj-he:
> LGTM, this prevents the error.
> 
> @aj-he should issues like this one and #1403 target a 1.0.1 patch, or do we want to prioritise 1.1.0? I suppose this opens up the wider question of what becomes AfHER LTS which we could discuss in the dev call.

@samuelscandrett - as 1.1 is not out yet, this bug should be patched in 1.0.x and then PR'd up into 1.1x. We have an incoming bug related to the referenced by scene that corrects functionality, and we'll be looking at doing that.

We have a change that adds ref-by scene to the Associated Resources tab in a number of resource type reports, this will be submitted to 1.1.x as it is an enhancement.

With respect to the LTS question, I think realistically we could only support it if it sits on an Arches LTS version.  We would need to be able to respond to Arches patches as quickly as possible.

samuelscandrett:
After discussions with @aj-he and @chiatt last week, 1.1 is the priority and all enhancements and fixes should go in there. We won't be working on a 1.0.1 patch, instead will be directing people to upgrade to 1.1.

PR #1415, opened 2026-01-28
Remove content-panel from consulation scss

### LINKED ISSUE #1414: [1.1 / 7.6] vertical scrollbar not showing in resource editor for cards with many widgets
On dev/1.1.x of AfHER, a card with many widgets such as "External Cross References" does not have a vertical scrollbar making it impossible to view all widgets and save/edit/delete the tile.

Image

After much debugging of the various AfHER CSS files, I have narrowed this bug down to this CSS property:

https://github.com/archesproject/arches-her/blob/e8a32690126d4a8d94f9438d27d9335a6f32c1c3/arches_her/media/css/consultations.scss#L358-L363

Rather than missing a scrollbar, it seems the page height is increased to 100%, tricking the front-end into thinking it doesn't need one. Commenting this CSS out reveals the scrollbar:

Image

samuelscandrett:
Closes #1414

### DIFF
--- a/arches_her/media/css/consultations.scss
+++ b/arches_her/media/css/consultations.scss
@@ -355,12 +355,6 @@ div.active-cons-header {
     align-items: center;
     display: flex;
 }
-.content-panel{
-    height: 100%;
-    .plugin-main{
-        height: 100%;
-    }
-}
 
 .active-cons-table-view {
     height: 57px;


phudson-he (approved):
LGTM! I cannot find an instance where these classes are being used any more, so must be "orphaned"

Issue #1389, opened 2025-10-10
White bars either side of Graphs/Branches in Arches Designer


samuelscandrett:
The list of Graphs/Branches in the Arches Designer page has white bars either side of each item. 
Presumably this is a CSS issue that occurred when moving from 7.5 to 7.6, as it was not AfHER 1.0.

<img width="1917" height="876" alt="Image" src="https://github.com/user-attachments/assets/a671685a-eb59-4b2e-8dd7-9ab02db15d87" />

samuelscandrett:
This looks to be caused by this CSS in `css/project-files/_landing.scss`

https://github.com/archesproject/arches-her/blob/d2e31c93288133ae5cb6061e2c8c7054f206593e/arches_her/media/css/project-files/_landing.scss#L5-L7

`$backgrounds` is imported from `css/_project-variables.scss`

https://github.com/archesproject/arches-her/blob/d2e31c93288133ae5cb6061e2c8c7054f206593e/arches_her/media/css/_project-variables.scss#L7

The issue here is the global affect of changing the background colour for all `body`. It is only called in one other place in [_landing.scss](https://github.com/archesproject/arches-her/blob/d2e31c93288133ae5cb6061e2c8c7054f206593e/arches_her/media/css/project-files/_landing.scss#L395), but it was mainly used for standardising the colour of the index.htm, as shown below:

| With body | Without body (slight change in background colour) |
| ----------- | ----------- |
| <img width="915" height="546" alt="Image" src="https://github.com/user-attachments/assets/2da9670d-778b-4209-bfdc-de0a7b4c28e1" />      | <img width="831" height="527" alt="Image" src="https://github.com/user-attachments/assets/c11377ab-15e9-44df-a9fd-800d80fbeba6" />       |
| <img width="1657" height="702" alt="Image" src="https://github.com/user-attachments/assets/3bffd400-15a9-4ecb-9ed0-97514ce2ba1a" />  | <img width="1462" height="658" alt="Image" src="https://github.com/user-attachments/assets/d238a973-a550-4d68-bfda-bef3d130424a" />      |





@phudson-he @CWDamm-Kint can we remove the global styling, affecting all body background colours, and only include it in the landing page scss instead?

phudson-he:
@samuelscandrett I see no harm in removing it and the white background works much better on the landing page, in my opinion. As long as it doesn't affect anything else.

CWDamm-Kint:
[...]

PR #1421, opened 2026-02-09
Move background colour setting from body to .landing-page

CWDamm-Kint:
As discussed in https://github.com/archesproject/arches-her/issues/1389, background colour was being set on body, so was impacting other parts of the site. 

We discussed whether removing this setting entirely looked better for the index page, but either way, this is something that should be controlled at the level of .landing-page, not body. So for now, I've just moved the setting into the landing-page class, which means we no longer see the erroneous white stripe on the other core Arches pages.


samuelscandrett:
LGTM!

PR #1425, opened 2026-02-10
Create number searching ES modifier

### LINKED ISSUE #1351: Numbers can't be searched in default search - AfHER workaround
In Arches 7, numbers can't be searched for in the default search.  This is/was because numbers weren't being added to the terms index.  This PR https://github.com/archesproject/arches/pull/11623 fixes the issue, but it is targeted at 8.0 due to being a new feature rather than a bug fix.

If we want this feature in the 7.6 version of AfHER, we could extend the index?

This issue is relevant to AfHER, as users should be able to search PRNs in the default search.  It is possible to do so in Advanced search, but that requires the knowledge of the resource model and is a few extra steps.

samuelscandrett:
Closes #1351

### DIFF
--- a/arches_her/settings.py
+++ b/arches_her/settings.py
@@ -74,6 +74,8 @@ ELASTICSEARCH_CUSTOM_INDEXES = []
 #     'should_update_asynchronously': False  <-- denotes if asynchronously updating the index would affect custom functionality within the project.
 # }]
 
+ES_MAPPING_MODIFIER_CLASSES = ["arches_her.utils.number_searching.NumberDataEsMappingModifier"]
+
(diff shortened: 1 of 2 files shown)


aj-he:
@samuelscandrett - FYI, we did a quick and dirty datatypes.py override to get PRNs searchable in GLHER. This may be adapted perhaps to be more closely aligned with your enhancement to v8?

https://github.com/HistoricEngland/arches-her/blob/keystone/release-1.3.x/arches_her/datatypes/datatypes.py

samuelscandrett:
Closing as this issue has been addressed in core 7.6 https://github.com/archesproject/arches/pull/12594

PR #1431, opened 2026-02-11
Remove docker support #1430

### LINKED ISSUE #1430: Remove Docker support as it does not currently work and is confusing for developers
All docker support to be removed as it is currently in a non working state that leads to confusion by users.

Instead we will direct users to use arches-containers for now while better container support is worked out.

aj-he:
- Removes aher_project docker component as it is not used in any other archesproject repo using the Arches application pattern. It also does not work at the moment.
- Removed the dockerfile and docker-compose as these are old and are not up to date.
- Updated the Readme to remove the docker instructions.

Closes #1430


chiatt (approved):
lgtm!

PR #1436, opened 2026-02-23
Fixes Referenced-by scene showing all relationships instead of just those that target the current resource

### LINKED ISSUE #1411: Referenced-by scene in resource reports has inaccuracies and is unclear in meaning to user.
**Reference-By**

 - It should show all resource types (not filter to a particular one like in Application Area which only shows Consultations), ensuring only reverse relationships are shown.

 - We need to be clear on what  "Referenced By" means, so to do this, we need to add a paragraph of text to describe the section as showing other resources that are linking to this resource

    _"This section shows other resources that hold a relationship to this resource."_

--- 
ergo:

To minimise Users confusion with the Resource Referenced By scene in resource reports:

 - Referenced-by to show clarifying statement to improve user feedback in the UI.

Image

Address issues found persuading Referenced-by to handle wildcard graphs and inaccuracies with potential two way relationships.

 - Referenced-by to filter by wildcard graphs and resource relationships **to** the current resource.

aj-he:
This pull request fixes the "Referenced By" report section by refining how related resources are filtered and displayed. The changes ensure that only resources which actually reference the current resource are shown.

It also adds a help note that tells the user that the lists shows resources that hold a relationship to this resource.

**Filtering and display improvements:**

* Updated the logic in `referenced-by.js` to filter related resources so that only those which point to the current resource are shown in the "Referenced By" section, using the `resource_relationships` data.
* Ensured that `self.graphsArray` is always initialized as an array to prevent errors when `params.graphs` is undefined.

**User interface enhancements:**

* Improved the conditional rendering in `referenced-by.htm` to show a clearer "no data" message when there are no referencing resources, and added a descriptive note explaining the section's purpose when resources are present.

Closes #1411

### DIFF
--- a/arches_her/media/js/views/components/reports/scenes/referenced-by.js
+++ b/arches_her/media/js/views/components/reports/scenes/referenced-by.js
@@ -20,6 +20,7 @@ function($, _, ko, arches, reportUtils, ReferencedByTemplate) {
             self.visible = {
                 referencedBy: ko.observable(true),
             };
+            self.loaderror = ko.observable(false);
 
             // referenced by table configuration
             self.referencedByTwoColumnTableConfig = {
@@ -43,12 +44,14 @@ function($, _, ko, arches, reportUtils, ReferencedByTemplate) {
                     }
                     return;
                 }).fail(function() {
-                    // error
+                    console.error("Unable to load graphs for referenced by scene");
+                    self.loaderror(true);
                 });
             };
 
 
-             self.getRelatedResources = function(){
+            self.getRelatedResources = function(){
+                self.loaderror(false);
                 return $.ajax({
                     url: arches.urls.related_resources + self.resourceinstanceid,
                     context: self,
@@ -83,7 +86,9 @@ function($, _, ko, arches, reportUtils, ReferencedByTemplate) {
                     });
                 })
                 .fail(function() {
-                    // error
+                    console.error("Unable to load related resources for referenced by scene");
+                    self.relations.removeAll();
+                    self.loaderror(true);
                 });
             };
 
--- a/arches_her/media/js/views/components/reports/scenes/referenced-by.js
+++ b/arches_her/media/js/views/components/reports/scenes/referenced-by.js
@@ -48,41 +48,45 @@ function($, _, ko, arches, reportUtils, ReferencedByTemplate) {
             };
 
 
-            self.getRelatedResources = function(){
+             self.getRelatedResources = function(){
                 return $.ajax({
                     url: arches.urls.related_resources + self.resourceinstanceid,
                     context: self,
                 })
-                    .done(function(response) {
-                        self.getGraphs().then(function(){
-                            let this_rid = response.related_resources.resource_instance.resourceinstanceid;
-                            //need to look at the resource_relationships and find other resources that are pointing TO this one
-                            other_resource_ids = [];
-                            for(const rr of response.related_resources.resource_relationships){
-                                if(rr.resourceinstanceidto == self.resourceinstanceid){
-                                    other_resource_ids.push(rr.resourceinstanceidfrom)
+                .done(function(response) {
+                    self.getGraphs().then(function(){
+                        const relationships = response.related_resources.resource_relationships;
+                        const relatedResources = response.related_resources.related_resources;
+                        // Collect resource IDs that point to this resource
+                        const otherResourceIds = relationships
+                            .filter(rr => rr.resourceinstanceidto === self.resourceinstanceid)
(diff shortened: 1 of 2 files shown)


arches_her/templates/views/components/reports/scenes/referenced-by.htm
```diff
@@ -8,13 +8,14 @@ <h2 class="aher-report-section-title"><i tabindex="0" data-bind="onEnterkeyClick
...
-    <!-- ko if: relations().length == 0 -->
+    <!-- ko ifnot: relations().length -->
     <div class="aher-nodata-note">{% trans "No resource referenced by for this resource" %}</div>
     <!-- /ko -->
 
-    <!-- ko if: relations().length > 0 -->
+    <!-- ko if: relations().length -->
     <div class="aher-report-subsection">
         <div class="firstchild-container">
+            <span class="aher-nodata-note">This section shows other resources that hold a relationship to this resource.</span>
```
samuelscandrett: Does this need wrapping in a translation template tag? 

```suggestion
            <span class="aher-nodata-note">{% trans "This section shows other resources that hold a relationship to this resource." %}</span>
```

arches_her/media/js/views/components/reports/scenes/referenced-by.js
```diff
@@ -48,30 +48,46 @@ function($, _, ko, arches, reportUtils, ReferencedByTemplate) {
...
                             }
-                            return;
                         });
-                    })
-                    .fail(function() {
-                        // error
                     });
+                })
+                .fail(function() {
+                    // error
```
samuelscandrett: Do we want to add anything here? 
Is a fail here severe enough for a `self.alert(new JsonErrorAlertViewModel('ep-alert-red', response.responseJSON));` or would a console log be sufficient?

samuelscandrett (changes requested):
Looks good, just a few minor suggestions/questions.

While testing I did spot a bug but it could be a core issue since I can replicate it on both branches. When relating a Consultation to an Application Area via the "Related Application Area" widget it shows up correctly in the "Resource Reference By" section (this is fine and expected behaviour). However, when you remove the relationship data from the tile but don't delete the tile this relationship remains in the "Resource Reference By" section. Only when removing the entire tile does the referenced relationship disappear.

The reason I think this is a core issue is because the removed relationship appears in both `response.related_resources.resource_relationships` and `response.related_resources.related_resources`.

Have you noticed this before @aj-he?

See the video below for a visual demonstration of the issue 

https://github.com/user-attachments/assets/c0230402-a185-4ee1-b21b-a6ae37356bba

aj-he:
> Looks good, just a few minor suggestions/questions.
> 
> While testing I did spot a bug but it could be a core issue since I can replicate it on both branches. When relating a Consultation to an Application Area via the "Related Application Area" widget it shows up correctly in the "Resource Reference By" section (this is fine and expected behaviour). However, when you remove the relationship data from the tile but don't delete the tile this relationship remains in the "Resource Reference By" section. Only when removing the entire tile does the referenced relationship disappear.
> 
> The reason I think this is a core issue is because the removed relationship appears in both `response.related_resources.resource_relationships` and `response.related_resources.related_resources`.
> 
> Have you noticed this before @aj-he?
> 
> See the video below for a visual demonstration of the issue
> 
>  20260224-1531-54.2222373.mp4

@samuelscandrett  This is a core problem from what I can tell. See my example below where the the application area is removed from the consultation and it is still present in the Related Resources list in the editor:

image

This shows the relationship is persisting externally of the ref-by scene.

arches_her/templates/views/components/reports/scenes/referenced-by.htm
aj-he: Good spot... applied

samuelscandrett:
> @samuelscandrett This is a core problem from what I can tell. See my example below where the the application area is removed from the consultation and it is still present in the Related Resources list in the editor:
> image
> 
> This shows the relationship is persisting externally of the ref-by scene.

Thanks @aj-he, sounds like a core bug then! I'll raise it over there

arches_her/media/js/views/components/reports/scenes/referenced-by.js
aj-he: @samuelscandrett - I've opted for a "failed to load" type response in the scene.  There is no alert binding available to the scene so thought this would do. Resolve if you're happy with this compromise?

samuelscandrett: Sounds good to me

samuelscandrett (approved):
LGTM!

PR #1441, opened 2026-02-25
Fixes Application Area MVT permissions Issue #1387

### LINKED ISSUE #1387: Application Areas MVT permission errors
The Application Areas MVT layer is failing In AfHER 7.6 due to changes in permissions backend `get_filtered_instances` (previously `get_filtered_resources`).
 https://github.com/archesproject/arches-her/blob/ea9659470281ec3cf9593433cf1e943b1b9f5bb2/arches_her/views/map.py#L18

[get_filtered_instances](https://github.com/archesproject/arches/blob/e49d6b60e6af782a480c37020c4728b55285d554/arches/app/utils/permission_backend.py#L201) now has a parameter for resources, which [get_filtered_resources](https://github.com/archesproject/arches/blob/05b5203fca96d37926183857a132bc05c7f7af21/arches/app/utils/permission_backend.py#L85) did not.

Do we need to change map.py to grab all App area resource ids and pass them in to `get_filtered_instances`?

aj-he:
This pull request updates the Application Area MVT (Mapbox Vector Tile) filtering logic to correctly handle both inclusive and exclusive resource instance lists, improving how map data is filtered based on user permissions. It also documents this fix in the release notes.

Improvements to MVT filtering logic:

* Updated the `get` method in `arches_her/views/map.py` to distinguish between exclusive and inclusive filtering modes using the return value from `get_filtered_instances`, ensuring the SQL query applies the correct inclusion or exclusion logic for resource instance IDs. [[1]](diffhunk://#diff-d3f53ca920820902c723c0abe3e7f87d297da0f2f48a3ee11fd737b7db94b862L18-R32) [[2]](diffhunk://#diff-d3f53ca920820902c723c0abe3e7f87d297da0f2f48a3ee11fd737b7db94b862L37-R43)

Documentation:

* Added a release note entry describing the fix for Application Area MVT filtering in `releases/1.1.0.md`.

### DIFF
--- a/arches_her/views/map.py
+++ b/arches_her/views/map.py
@@ -15,16 +15,21 @@ class ApplicationAreas(View):
         try:
             node = models.Node.objects.get(nodeid=nodeid, nodegroup_id__in=viewable_nodegroups)
             se = SearchEngineFactory().create()
-            restricted_resource_ids = get_filtered_instances(request.user, search_engine=se)
-            if len(restricted_resource_ids) == 0:
-                restricted_resource_ids.append(
+            #restricted_resource_ids = get_filtered_instances(request.user, search_engine=se)
+            
+            # get_filtered_instances returns a list of ids and states whether they are exclusive (exclude the resourceids - from default deny)     
+            # or inclusive (only include these resourceids - from default allow).
+            is_exclusive, filtered_instances = get_filtered_instances(request.user, search_engine=se)
+
+            if len(filtered_instances) == 0:
+                filtered_instances.append(
                     "10000000-0000-0000-0000-000000000001"
-                )  # This must have a uuid that will never be a resource id.
-            restricted_resource_ids = tuple(restricted_resource_ids)
+                )  # This must have a uuid that will likely never be a resource id so to not break the SQL query when there are no resources to filter on.
+            filtered_instances = tuple(filtered_instances)
 
             with connection.cursor() as cursor:
                 result = cursor.execute(
-                    """SELECT ST_AsMVT(tile, 'app-area', 4096, 'geom', 'id') FROM (SELECT tileid,
+                    f"""SELECT ST_AsMVT(tile, 'app-area', 4096, 'geom', 'id') FROM (SELECT tileid,
                         id,
                         resourceinstanceid,
                         nodeid,
@@ -34,8 +39,8 @@ class ApplicationAreas(View):
                         ) AS geom,
                         1 AS total
                     FROM geojson_geometries
-                    WHERE nodeid = %s and resourceinstanceid not in %s and (geom && ST_TileEnvelope(%s, %s, %s))) AS tile;""",
-                    [zoom, x, y, nodeid, restricted_resource_ids, zoom, x, y],
+                    WHERE nodeid = %s and resourceinstanceid {'' if is_exclusive else 'not'} in %s and (geom && ST_TileEnvelope(%s, %s, %s))) AS tile;""",
+                    [zoom, x, y, nodeid, filtered_instances, zoom, x, y],
                 )
                 result = bytes(cursor.fetchone()[0]) if result is None else result
             return HttpResponse(result, content_type="application/x-protobuf")
--- a/arches_her/views/map.py
+++ b/arches_her/views/map.py
@@ -18,12 +18,12 @@ class ApplicationAreas(View):
 
             if node is None:
                 return HttpResponse(status=503)
-            
+
             candidate_resource_ids = [
                 str(resourceinstanceid)
-                for resourceinstanceid in models.ResourceInstance.objects.filter(
-                    graph_id=node.graph_id
-                ).values_list("resourceinstanceid", flat=True)
+                for resourceinstanceid in models.ResourceInstance.objects.filter(graph_id=node.graph_id).values_list(
+                    "resourceinstanceid", flat=True
+                )
             ]
 
             resource_filter_sql = ""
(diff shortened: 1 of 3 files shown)


samuelscandrett:
Thanks @aj-he. @CWDamm-Kint was looking into creating a test to cover this issue, but since this branch is on the HE fork and we'd probably have to mess around with permissions, I think it is be sensible to follow up with another PR.
I raised an issue for it here https://github.com/archesproject/arches-her/issues/1444

arches_her/views/map.py
```diff
@@ -15,16 +15,20 @@ def get(self, request, zoom, x, y):
         try:
             node = models.Node.objects.get(nodeid=nodeid, nodegroup_id__in=viewable_nodegroups)
             se = SearchEngineFactory().create()
-            restricted_resource_ids = get_filtered_instances(request.user, search_engine=se)
-            if len(restricted_resource_ids) == 0:
-                restricted_resource_ids.append(
+
+            # get_filtered_instances returns a list of ids and states whether they are exclusive (exclude the resourceids - from default deny)
+            # or inclusive (only include these resourceids - from default allow).
+            is_exclusive, filtered_instances = get_filtered_instances(request.user, search_engine=se)
```
CWDamm-Kint: As above, I think this currently needs a list of all resource ids as strings:

`all_resources = [str(id) for id in Resource.objects.values_list('resourceinstanceid', flat=True)]`

`is_exclusive, filtered_instances = get_filtered_instances(request.user, search_engine=se, resources=all_resources)`

CWDamm-Kint (changes requested):
Hi @aj-he. I've tested this and I'm afraid it still seems to give me 500 errors as a non-super user. I added an overlay layer based on the MVT source, to be able to test this easily with additional users, who may not have immediate access to the consultation workflows (thanks @samuelscandrett for this idea). 

To replicate, the layer definition is `[{"id": "select-application-area", "type": "fill", "paint": {"fill-color": "green", "fill-opacity": 0.25, "fill-outline-color": "black"}, "source": "select-application-area", "source-layer": "app-area"}]`. I've also got the formal test @samuelscandrett mentioned, which I'll upload as a separate branch shortly. 

I've added what worked for me below, but arguably the real issue is in core's `get_filtered_instances` in `utils/permission_backend`, as it has the type suggestion set as the default parameter. If not for this, I don't think you'd need to provide a resources argument. I'll raise that as a separate issue, but the suggested workaround might be ok for now, given 1.1.0 timescales?

Edit: Core issue raised: https://github.com/archesproject/arches/issues/12618

arches_her/views/map.py
aj-he: @CWDamm-Kint - yes, you were correct and it now fetches resource instances that users have permissions to access. I did a few other tweaks while I was there as it was bugging me a little.

aj-he:
@samuelscandrett @CWDamm-Kint - I've added some basic tests for this. I wanted to submit at least something as part of this PR.

The compromise is that I have patched the `get_filtered_instances` so that the response is not dependant on my spending time creating test data and permissions specifically for the deny and allow backends.

Tbh, I don't think it is for this PR to ensure that the backends are returning the right data - rather we should assume this is being tested separately.

If you want to do this in #1444 then I'll leave that up to you both.

CWDamm-Kint (approved):
Thanks @aj-he. The functionality is working for me now, as a super user and a regular user. It's also passing the tests that I wrote. 

For the tests in this PR, I'm assuming the plan is ultimately to add assertions that the right ids are in the `response.content`? But happy for now if you just want to merge these as they are.

aj-he:
> Thanks @aj-he. The functionality is working for me now, as a super user and a regular user. It's also passing the tests that I wrote.
> 
> For the tests in this PR, I'm assuming the plan is ultimately to add assertions that the right ids are in the `response.content`? But happy for now if you just want to merge these as they are.

@CWDamm-Kint - Do you mean check the contents of the `results` (https://github.com/HistoricEngland/arches-her/blob/ed67caf4e05116ff659c4123edf7dd10f867958b/arches_her/views/map.py#L67) to ensure that the resource UUIDs are present in the response?

image

I had excluded this idea as PSQL sends back a bytea value to somehow decode into a meaningful object etc, but it appears that converting it to a string allows you to quite easily regex the UUIDs and check the resourceinstanceid UUIDs are present.

I've pushed an update with this in.

CWDamm-Kint:
> @CWDamm-Kint - Do you mean check the contents of the `results` (https://github.com/HistoricEngland/arches-her/blob/ed67caf4e05116ff659c4123edf7dd10f867958b/arches_her/views/map.py#L67) to ensure that the resource UUIDs are present in the response?

@aj-he - yes that's right. I think you can also import `mapbox_vector_tile`, and use `decode` to make it a bit more human readable: `decoded_data = mapbox_vector_tile.decode(response.content)`, but you're right that casting the bytea into a string also works. I was actually originally thinking something lazier and just checking whether the id string was found anywhere in the converted string, but extracting them with regex is definitely a more precise option!

CWDamm-Kint:
For reference I've added my work to this draft PR: https://github.com/archesproject/arches-her/pull/1448

I think it's probably duplicative now, but maybe have a look @aj-he and let me know if you think the tests would still be worth incorporating somehow.

aj-he:
> > @CWDamm-Kint - Do you mean check the contents of the `results` (https://github.com/HistoricEngland/arches-her/blob/ed67caf4e05116ff659c4123edf7dd10f867958b/arches_her/views/map.py#L67) to ensure that the resource UUIDs are present in the response?
> 
> @aj-he - yes that's right. I think you can also import `mapbox_vector_tile`, and use `decode` to make it a bit more human readable: `decoded_data = mapbox_vector_tile.decode(response.content)`, but you're right that casting the bytea into a string also works. I was actually originally thinking something lazier and just checking whether the id string was found anywhere in the converted string, but extracting them with regex is definitely a more precise option!

@CWDamm-Kint - `mapbox-vector-tile` was removed at arches v7.6.0 so I didn't want to reinstate it.

> For reference I've added my work to this draft PR: https://github.com/archesproject/arches-her/pull/1448
>
> I think it's probably duplicative now, but maybe have a look @aj-he and let me know if you think the tests would still be worth incorporating somehow.

Your tests are probably more representative of real world as you are loading graphs, ontologies and curated resources.

We may want to hold fire on yours as v2.0.0 will likely make significant changes to graphs etc as it adopts v8.1/2.

PR #1447, opened 2026-03-02
Adds migration for workflow config, help button and starting help templates

### LINKED ISSUE #1443: Move workflow config and add help templates
As a follow up to #1442, we should introduce a new migration file for moving individual workflow config data from init-workflow into the workflows themselves.

This will allow a source of truth per workflow, and will enable us to add help templates.

samuelscandrett:
Closes #1443 

A couple of important points:
- This branch extends the existing 0004 migration rather than creating a new one, since 0004 refers to altering plugins and it seemed well placed there. To best test this PR:
   1. Unmigrate to 0003 on dev/1.1.x 
       `python manage.py migrate arches_her 0003`
   2. Fetch and checkout this git branch
   3. Migrate 0004 again
      `python manage.py migrate arches_her`
- The migration includes forward and backward for both individual and init-workflow config and help templates. 
For whatever reason, it doesn't seem possible to remove the `workflows: []` property from the object via python, so I've had to leave it empty.
- I've added some simple words to the help templates. The Communication and Correspondence help is currently unfinished due to current errors in the workflows preventing me from progressing and seeing which fields are required/optional.
- In AfS, the help menu uses the custom KoJS binding "slide", however I've used a flexbox layout to fix an issue on AfS where help information was showing in a small box and this has resulted in slide failing to work since it needs a fixed height. As a result, I've removed the slide animation, instead the window just reveals and hides as normal.
- The help templates are using an `iframe` which is blocked by default. I've added a line into settings.py [here](https://github.com/archesproject/arches-her/pull/1447/changes#diff-4fc5f6370d3adeb869cd4c088c851462a116f288417173addd846c21569df8aaR365) to fix this, but let me know if it shouldn't be there or should live elsewhere.

### DIFF
--- a/arches_her/settings.py
+++ b/arches_her/settings.py
@@ -362,7 +362,6 @@ RENDERERS = [
     },
 ]
 
-X_FRAME_OPTIONS = "SAMEORIGIN"
 
 # By setting RESTRICT_MEDIA_ACCESS to True, media file requests outside of Arches will checked against nodegroup permissions.
 RESTRICT_MEDIA_ACCESS = False
--- a/arches_her/settings.py
+++ b/arches_her/settings.py
@@ -362,6 +362,8 @@ RENDERERS = [
     },
 ]
 
+X_FRAME_OPTIONS = "SAMEORIGIN"
+
 # By setting RESTRICT_MEDIA_ACCESS to True, media file requests outside of Arches will checked against nodegroup permissions.
 RESTRICT_MEDIA_ACCESS = False
 
--- a/arches_her/templates/views/components/plugins/init-workflow.htm
+++ b/arches_her/templates/views/components/plugins/init-workflow.htm
@@ -78,6 +78,60 @@
                 {% trans 'Show Incomplete Workflows' %}
             </button>
 
+            <!-- Workflow Help button -->
+            <button 
+                class="btn btn-info btn-sm"
+                style="margin: 5px; margin-right: 10px; margin-left: 0px; border-radius: 4px;"
+                data-bind="
+                    click: function(){ $data.shouldShowWorkflowHelp(true); }
+                "
+            >
+                {% trans 'Show Workflow Help' %}
+            </button>
+        </div>
+
+        <div id="ep-help-panel" 
+            class="ep-help" 
+            style="display: none"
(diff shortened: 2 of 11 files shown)


arches_her/settings.py
```diff
@@ -362,6 +362,8 @@
     },
 ]
 
+X_FRAME_OPTIONS = "SAMEORIGIN"
```
chiatt: I think this is okay here for this PR, however, we should really find an alternative to using an iframe (probably used in afs as well). Something like this should work: https://github.com/archesproject/arches-her/pull/1450/changes. That change could go in a patch if necessary.

chiatt (approved):
I think this looks good, however, you might want to merge this little refactor first: https://github.com/archesproject/arches-her/pull/1449.

arches_her/settings.py
samuelscandrett: I agree and thought the same about an iframe replacement whilst working on the PR. Thanks for raising a PR, I'll look into it

samuelscandrett: Resolved as #1450 now merged

samuelscandrett:
Following improvements from #1450 and #1449, I think this might be ready to go?

chiatt (approved):
lgtm!

aj-he (approved):
works for me, ship it.

arches_her/templates/views/components/plugins/init-workflow.htm
```diff
@@ -68,36 +72,96 @@ <h2>{% trans 'Incomplete Workflows' %}</h2>
...
+                                dropdownCssClass: 'select2-zindex',
+                                closeOnSelect: true,
+                                allowClear: true,
+                                multiple: false,
+                                placeholder: '{% trans "Select A Workflow" %}'
+                            }
+                        }
+                    "
+                ></select>
+            </div>
```
samuelscandrett, line 154: @aj-he Does this pass accessibility as is? Tabbing through seems to work, and I've used NVDA to check screen reader output. Although it's missing `attr: {'data-label': ...}`, NVDA does read it out as "application area select a workflow combo box collapsed" (not sure why it calls it application area, presumably grabbing the first in the list from somewhere)

aj-he: See https://github.com/archesproject/arches-her/pull/1451 where I have suggested some general WCAG 2.2 AA changes.

Issue #1445, opened 2026-02-26
Tabbing through the incomplete workflows modal also highlights the workflow buttons behind


samuelscandrett:
Identified in #1442

When tabbing through the incomplete workflows modal menu, the workflow buttons are also highlighted. This shouldn't be the case as they can't be clicked normally until the modal menu is closed.

https://github.com/user-attachments/assets/e1d234b1-a841-43ac-ae06-543bb3630904

aj-he:
#1443 includes fix for keyboard tabbing moving focus outside of model

PR #1451, opened 2026-03-09
[review only] workflows page additional accessibility changes

aj-he:
@samuelscandrett - there were a few things that came up when I took a look with respect to accessibility so I thought I'd send you a PR of those changes which you can merge if you like.

#1443

### DIFF
--- a/arches_her/media/js/views/components/plugins/init-workflow.js
+++ b/arches_her/media/js/views/components/plugins/init-workflow.js
@@ -68,6 +68,19 @@ define([
         this.shouldShowIncompleteWorkflowsModal = ko.observable(false);
         this.requestingUserIsSuperuser = ko.observable(false);
 
+        // Subscribe to modal state to manage focus with delay for screen readers
+        this.shouldShowIncompleteWorkflowsModal.subscribe(isOpen => {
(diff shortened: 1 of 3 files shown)


arches_her/media/js/views/components/plugins/init-workflow.js
```diff
@@ -68,6 +68,19 @@ define([
...
+                // Delay focus to allow screen readers to announce dialog context
+                setTimeout(() => {
+                    const closeButton = document.getElementById('close-workflow-modal-btn');
+                    if (closeButton) {
+                        closeButton.focus();
+                    }
+                }, 100);
+            }
+        });
+
```
samuelscandrett, line 83: I like this, and it solves #1445.

One difference I notice here is that with the incomplete wf modal open I can't tab to the navbar, only to the links and the close button. Are we happy with this as the intended behaviour? I'm not opposed to it.

samuelscandrett (approved):
LGTM! 

Happy to approve, just one small comment.

I notice that the help combobox still reads "application area" at the start which is odd. I don't suppose this is blocking this PR though, and possibly could be raised separately.

PR #1464, opened 2026-04-23
1462 Add Checks for Pre-existing Extensions in initial migration

### LINKED ISSUE #1462: Add checks to initial migration file
When migrating a pre-existing database with BNG functions and datatype already registered, the migrations fail due to duplicate primary keys.

A check needs to be put in place to stop this issue from happening

csmith-he:
# Pull Request Template

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - [ ] Involves changes to the Model/Graph, Concepts or Collections, or ontologies.

## Description of Change

Adds an .exists() check to migrations for extensions and skips them if they exist so that an error message relating to primary keys is not returned

## Issues Solved

Closes #1462

## Checklist

  - [X] `dev/1.1.x` (default): features and bugfixes for the current release line

## Ticket Background

- Sponsored by: Historic England
- Found by: @csmith-he 
- Tested by: @csmith-he 
- Designed by: @csmith-he

### DIFF
--- a/arches_her/migrations/0001_initial.py
+++ b/arches_her/migrations/0001_initial.py
@@ -20,142 +20,150 @@ class Migration(migrations.Migration):
         Plugin = apps.get_model("models", "Plugin")
 
         # Add Accessibility
-        Plugin.objects.update_or_create(
-            pluginid="8688eaf9-3605-4384-823c-d01c0d210f82",
-            name="Accessibility",
-            icon="fa fa-universal-access",
-            component="views/components/plugins/accessibility",
-            componentname="accessibility",
-            slug="accessibility",
-            config={},
-            sortorder=0,
-        )
+        if not Plugin.objects.filter(pk="8688eaf9-3605-4384-823c-d01c0d210f82").exists():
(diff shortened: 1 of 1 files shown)


aj-he (approved):
Looks good WRT handling existing UUIDs. I don't have a good 7.5 environment to do a test run, but I know you have doing this with a dataset, so happy to accept.

samuelscandrett (changes requested):
I tested this by removing all arches_her migrations (by removing them from the django_migrations table rather than zeroing which prevents it).

One error I did still get was:
```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "map_sources_name_key"
DETAIL:  Key (name)=(select-application-area) already exists.
```

After deleting this from the db migrations were able to run properly, so I think we need to wrap this too

samuelscandrett (approved):
Looks good to me!

PR #1468, opened 2026-04-30
Update min arches version and package-lock.json

chiatt:
## Description of Change

Updates min arches version to 7.6.23 and package-lock.json in advance of 1.1.0 release


aarongundel (approved):
lgtm

PR #1469, opened 2026-04-30
Fixes to support build

chiatt:
1. Update build manifest to exclude files we don't want in the build
2. gitignore dist directory
3. Fix license formatting


aarongundel (approved):
LGTM!

PR #1472, opened 2026-05-05
1466 only save sys ref tile when modified

### LINKED ISSUE #1466: GenerateUniqueReferences saves the system ref tile, even when it remains unaltered
In `generate_unique_references_function.py`, `check_and_populate_uids` returns TRUE, regardless of whether the id variables have been modified. This means that the previously saved reference tile gets resaved, which adds quite a lot of overhead to `tile.save()`, as the resource reindex is triggered twice.

### LINKED ISSUE #1471: The function to generate unique references runs a UUID check on an internationalised string.
The helper function, `check_and_populate_uuids`, within `generate_unique_references_function.py`, calls a check on an internationalised string. This is the resource id stored on the system references tile:

```
if currentTile.data[resid_node] is not None:
                        try:
                            UUID(currentTile.data[resid_node])
                            pass
                        except:
                            currentTile.data[resid_node] = format_string_value(str(resourceidval))
                    else:
                        currentTile.data[resid_node] = format_string_value(str(resourceidval))
```
This means that it always currently fails, leading to the except block and overwriting the resource id string.

CWDamm-Kint:
# Pull Request Template

This PR ensures that an existing system reference tile is only resaved (and hence re-indexed), if changes are made. 

It also rewrites the check for a valid resource id on that tile, to ensure it works with internationalised strings. 

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)

## Description of Change

A `has_changes` variable is created, which is set to `True` if either `populate_simple_id` or the newly created `populate_resid_id` are successfully called. The `check_and_populate_uids` then returns the `has_changes` value, rather than just automatically returning `True`. 

## Issues Solved

Closes https://github.com/archesproject/arches-her/issues/1466
Closes https://github.com/archesproject/arches-her/issues/1471

## Checklist

  - [x] `dev/1.1.x` (default): features and bugfixes for the current release line

## Ticket Background

- Sponsored by: Knowledge Integration
- Found by: @CWDamm-Kint 
- Tested by: @CWDamm-Kint 
- Designed by: @CWDamm-Kint

### DIFF
--- a/arches_her/functions/generate_unique_references_function.py
+++ b/arches_her/functions/generate_unique_references_function.py
@@ -124,25 +124,21 @@ class GenerateUniqueReferences(BaseFunction):
 
             def simpleid_nextval_table_exists():
                 with connection.cursor() as cursor:
-                    cursor.execute(
-                        """
+                    cursor.execute("""
                         SELECT EXISTS (
                             SELECT FROM information_schema.tables 
                             WHERE  table_schema = 'public'
                             AND    table_name   = 'simpleid_nextval'
                         );
-                    """
-                    )
+                    """)
                     return bool(cursor.fetchone()[0])
 
             def get_next_simpleid():
                 with connection.cursor() as cursor:
-                    cursor.execute(
-                        """
+                    cursor.execute("""
                         TRUNCATE TABLE simpleid_nextval;
                         INSERT INTO simpleid_nextval DEFAULT VALUES RETURNING id;
-                    """
-                    )
+                    """)
                     return int(cursor.fetchone()[0])
 
             def get_simple_id_nodeinfo(functionid=None):
@@ -187,6 +183,9 @@ class GenerateUniqueReferences(BaseFunction):
                     currentTile.data[resid_node_id] = {language.code: {"value": resourceid_val, "direction": language.default_direction}}
                     return True
 
+                def get_formatted_id(id_string, language):
+                    return id_string[language.code]["value"]
+
                 try:
                     has_changes = False
                     simpleid = currentTile.data[simpleid_node]
@@ -205,7 +204,8 @@ class GenerateUniqueReferences(BaseFunction):
 
                     if currentTile.data[resid_node] is not None:
                         try:
-                            UUID(currentTile.data[resid_node])
+                            resid_string = get_formatted_id(currentTile.data[resid_node], default_language)
+                            UUID(resid_string)
                         except:
                             has_changes = populate_resid_id(currentTile, resid_node, resourceidval, default_language)
                     else:
--- a/arches_her/functions/generate_unique_references_function.py
+++ b/arches_her/functions/generate_unique_references_function.py
@@ -214,14 +214,14 @@ class GenerateUniqueReferences(BaseFunction):
                         try:
                             x = int(currentTile.data[simpleid_node])
                             self.logger.debug("Resource " + str(resourceidval) + "has valid simpleid: " + str(x))
-                        except:
+                        except (ValueError, TypeError):
                             has_changes = populate_simple_id(currentTile, simpleid_node)
 
                     if currentTile.data[resid_node] is not None:
                         try:
                             resid_string = get_formatted_id(currentTile.data[resid_node], default_language)
                             UUID(resid_string)
-                        except:
+                        except (ValueError, TypeError):
                             has_changes = populate_resid_id(currentTile, resid_node, resourceidval, default_language)
                     else:
                         has_changes = populate_resid_id(currentTile, resid_node, resourceidval, default_language)
--- a/arches_her/functions/generate_unique_references_function.py
+++ b/arches_her/functions/generate_unique_references_function.py
@@ -178,6 +178,7 @@ class GenerateUniqueReferences(BaseFunction):
                 def populate_simple_id(currentTile, simple_node_id):
                     nextsimpleval = get_next_simple_id()
                     currentTile.data[simple_node_id] = nextsimpleval
+                    return True
 
                 def populate_resid_id(currentTile, resid_node_id, resourceid_val, language):
                     currentTile.data[resid_node_id] = {language.code: {"value": resourceid_val, "direction": language.default_direction}}
--- a/arches_her/functions/generate_unique_references_function.py
+++ b/arches_her/functions/generate_unique_references_function.py
@@ -198,7 +198,7 @@ class GenerateUniqueReferences(BaseFunction):
 
                     default_language = models.Language.objects.get(code=settings.LANGUAGE_CODE)
 
-                    if currentTile.data.get(simpleid_node):
+                    if currentTile.data[simpleid_node] is not None and currentTile.data[simpleid_node] != 0:
                         try:
                             x = int(currentTile.data[simpleid_node])
                             self.logger.debug(f"Resource {resourceidval} has valid simpleid: {x}")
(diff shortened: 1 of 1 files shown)


arches_her/functions/generate_unique_references_function.py
```diff
@@ -182,36 +182,43 @@ def check_and_populate_uids(currentTile, simpleid_node, resid_node, resourceidva
...
+                        language.code: {"value": str(resourceid_val), "direction": language.default_direction}
+                    }
+                    return True
+
+                def get_formatted_id(id_string, language):
+                    return id_string[language.code]["value"]
+
+                try:
+                    has_changes = False
+                    simpleid = currentTile.data[simpleid_node]
```
samuelscandrett: I'd suggest to remove the creation of `simpleid` variable, purely because `currentTile.data[simpleid_node]` is still in use further down this file, so it becomes confusing dealing with two variables of the same values. Additionally, if anything else changes in `currentTile.data` it might not be reflected back in `simpleid` so there's potential data loss

arches_her/functions/generate_unique_references_function.py
```diff
@@ -182,36 +182,43 @@ def check_and_populate_uids(currentTile, simpleid_node, resid_node, resourceidva
...
-                        if currentTile.data[simpleid_node] != 0:
-                            try:
-                                x = int(currentTile.data[simpleid_node])
-                                self.logger.debug("Resource " + str(resourceidval) + "has valid simpleid: " + str(x))
-                                pass
-                            except:
-                                populate_simple_id(currentTile, simpleid_node)
-                        else:
-                            populate_simple_id(currentTile, simpleid_node)
+                    if simpleid is None or simpleid == 0:
```
samuelscandrett: I like the combination of the previous ifs on lines 192 and 193, but is there a reason you've flipped the condition logic? 

Would it be simpler to keep the logic as 
```
if currentTile.data[simpleid_node] is not None or currentTile.data[simpleid_node] != 0:
    try:
        x = int(currentTile.data[simpleid_node])
        self.logger.debug("Resource " + str(resourceidval) + "has valid simpleid: " + str(x))
        pass
    except:
        has_changes = populate_simple_id(currentTile, simpleid_node)                        
else:
    has_changes = populate_simple_id(currentTile, simpleid_node)
```

arches_her/functions/generate_unique_references_function.py
```diff
@@ -182,36 +182,43 @@ def check_and_populate_uids(currentTile, simpleid_node, resid_node, resourceidva
...
                     default_language = languages.get(code=settings.LANGUAGE_CODE)
-                    return {default_language.code: {"value": id_string_value, "direction": default_language.default_direction}}
 
-                try:
-                    if currentTile.data[simpleid_node] is not None:
-                        if currentTile.data[simpleid_node] != 0:
-                            try:
-                                x = int(currentTile.data[simpleid_node])
-                                self.logger.debug("Resource " + str(resourceidval) + "has valid simpleid: " + str(x))
-                                pass
```
samuelscandrett, line 197: we've also lost this `pass`

arches_her/functions/generate_unique_references_function.py
```diff
@@ -182,36 +182,43 @@ def check_and_populate_uids(currentTile, simpleid_node, resid_node, resourceidva
...
+                        except:
+                            has_changes = populate_simple_id(currentTile, simpleid_node)
 
                     if currentTile.data[resid_node] is not None:
                         try:
-                            UUID(currentTile.data[resid_node])
-                            pass
+                            resid_string = get_formatted_id(currentTile.data[resid_node], default_language)
+                            UUID(resid_string)
                         except:
```
samuelscandrett: What does this bit actually do? I see we're creating `resid_string` (which is just the resourceinstanceid) and assigning it the type UUID, but it's never used again after this. 
Commenting it out doesn't seem to break it, or at least not that I can see!

arches_her/functions/generate_unique_references_function.py
```diff
@@ -182,36 +182,43 @@ def check_and_populate_uids(currentTile, simpleid_node, resid_node, resourceidva
...
+                def get_formatted_id(id_string, language):
+                    return id_string[language.code]["value"]
+
+                try:
+                    has_changes = False
+                    simpleid = currentTile.data[simpleid_node]
 
-                def format_string_value(id_string_value):
                     languages = models.Language.objects.all()
                     default_language = languages.get(code=settings.LANGUAGE_CODE)
```
samuelscandrett: Here we're getting all language objects every time, and while I know it's not a hefty call to the db (as people aren't likely to have more than 10 languages loaded) it could be reduced by just using `default_language = models.Language.objects.get(code=settings.LANGUAGE_CODE)`

samuelscandrett (changes requested):
Looking good, and some nice fixes in here! 
Just a few comments & suggested changes

arches_her/functions/generate_unique_references_function.py
```diff
@@ -182,36 +182,43 @@ def check_and_populate_uids(currentTile, simpleid_node, resid_node, resourceidva
...
+                        try:
+                            x = int(currentTile.data[simpleid_node])
+                            self.logger.debug("Resource " + str(resourceidval) + "has valid simpleid: " + str(x))
+                        except:
+                            has_changes = populate_simple_id(currentTile, simpleid_node)
 
                     if currentTile.data[resid_node] is not None:
                         try:
-                            UUID(currentTile.data[resid_node])
-                            pass
```
samuelscandrett, line 208: and this pass too

arches_her/functions/generate_unique_references_function.py
CWDamm-Kint: I think pass is only needed when no other expression is present. Kind of a placeholder.

arches_her/functions/generate_unique_references_function.py
CWDamm-Kint: I think it's a validation check. `UUID(reside_string)` isn't actually assigned to anything. So if the resource id string is invalid, then it trips the except block. I've added a the exception types though (for if the UUID is invalid, or None).

arches_her/functions/generate_unique_references_function.py
CWDamm-Kint: I've adopted yours, but reduced the first 'if check' to just check whether it's falsey (as None and 0 both are). I also ditched the pass, as mentioned above. See https://github.com/archesproject/arches-her/pull/1472/changes/ef71e2b20aa93886c8f9f36f1ed8553cc2d809d9

samuelscandrett (approved):
LGTM

@aj-he what do you think? I know we plan to remove these AfHER functions in favour of the HE extension apps, but for now whilst they're a feature of AfHER I think it'd be wise to patch them

aj-he:
> LGTM
> 
> @aj-he what do you think? I know we plan to remove these AfHER functions in favour of the HE extension apps, but for now whilst they're a feature of AfHER I think it'd be wise to patch them

@samuelscandrett - Yes, keep changes in afher for now until we've discuss with the tech committee. I think we are adding this function to our arches-he-sysref-func extension now so I'll ask the team to look at this.

samuelscandrett:
> @samuelscandrett - Yes, keep changes in afher for now until we've discuss with the tech committee. I think we are adding this function to our arches-he-sysref-func extension now so I'll ask the team to look at this.

Thanks @aj-he, sounds good to me. I'll merge this in then. 
I think we should start to consider releasing 1.1.1, especially since this bugfix helps to improve performance too.

PR #1473, opened 2026-05-05
1465 add refresh_tile_geojson_geometries to BNGPointToGeoJSON function

### LINKED ISSUE #1465: Refresh_geojson_geometries slows down BNGPointToGeoJSON
Running the BNGPointToGeoJSON function (when creating GeoJSON from an entered BNG value), takes 40-50 seconds with approximately 300k rows in the geojson_geometries table. 

We've narrowed down the bottleneck to the `refresh_geojson_geometries` PSQL function here:

https://github.com/archesproject/arches-her/blob/6e3b2bacc131c2007d02c2b88cd9742f82002c9c/arches_her/functions/bngpoint_to_geojson_function.py#L211-L215

https://github.com/user-attachments/assets/315e97a0-9609-4c4b-9b19-114669a0cf3a

CWDamm-Kint:
# Pull Request Template

`refresh_geojson_geometries` has been creating a bottleneck during saving in the BNGPointToGeoJSON function, when there are a large number of resources with geometries. As described in https://github.com/archesproject/arches-her/issues/1465, there is an alternative that only updates the geometry for the individual tile. 

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - [ ] Involves changes to the Model/Graph, Concepts or Collections, or ontologies.

## Issues Solved

Closes https://github.com/archesproject/arches-her/issues/1465

## Checklist

  - [x] `dev/1.1.x` (default): features and bugfixes for the current release line

## Ticket Background

- Sponsored by: Knowledge Integration
- Found by: @CWDamm-Kint 
- Tested by: @CWDamm-Kint 
- Designed by: @CWDamm-Kint

### DIFF
--- a/arches_her/functions/bngpoint_to_geojson_function.py
+++ b/arches_her/functions/bngpoint_to_geojson_function.py
@@ -184,6 +184,7 @@ class BNGPointToGeoJSON(BaseFunction):
 
             if self.config["geojson_nodegroup"] == str(tile.nodegroup_id):
                 tile.data[geojsonNode] = geometryValueJson
+                tile.save()
                 tileid_to_refresh = tile.tileid
             else:
                 previously_saved_tiles = Tile.objects.filter(
--- a/arches_her/functions/bngpoint_to_geojson_function.py
+++ b/arches_her/functions/bngpoint_to_geojson_function.py
@@ -209,10 +209,11 @@ class BNGPointToGeoJSON(BaseFunction):
                     new_geojson_tile.save()
 
             cursor = connection.cursor()
-            sql = """
-                    SELECT * FROM refresh_tile_geojson_geometries(%s);
-                """
-            cursor.execute(sql, (tile.tileid,))
+
+            cursor.execute(
+                "SELECT * FROM refresh_tile_geojson_geometries(%s);",
+                (tile.tileid,),
+            )
 
         else:
             pass
--- a/arches_her/functions/bngpoint_to_geojson_function.py
+++ b/arches_her/functions/bngpoint_to_geojson_function.py
@@ -180,8 +180,11 @@ class BNGPointToGeoJSON(BaseFunction):
             The new tile is saved and then the mv_geojson_geoms materialised view is refreshed so the point geometry will be displayed
             on the Search map.
             """
+            tileid_to_refresh = None
+
             if self.config["geojson_nodegroup"] == str(tile.nodegroup_id):
                 tile.data[geojsonNode] = geometryValueJson
+                tileid_to_refresh = tile.tileid
             else:
                 previously_saved_tiles = Tile.objects.filter(
                     nodegroup_id=self.config["geojson_nodegroup"], resourceinstance_id=tile.resourceinstance_id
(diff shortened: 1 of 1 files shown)


arches_her/functions/bngpoint_to_geojson_function.py
```diff
@@ -210,9 +210,9 @@ def save_geojson(self, tile, request, is_function_save_method=True):
 
             cursor = connection.cursor()
             sql = """
-                    SELECT * FROM refresh_geojson_geometries();
+                    SELECT * FROM refresh_tile_geojson_geometries(%s);
                 """
-            cursor.execute(sql)  #
+            cursor.execute(sql, (tile.tileid,))
```
samuelscandrett: Wondering whether we could include the s string within the sql query string rather than the execute function, like in these two examples from core?

https://github.com/archesproject/arches/blob/381ed0068f69fe040ee5ecf308c686f3e96202bd/arches/app/etl_modules/save.py#L46-L50

https://github.com/archesproject/arches/blob/381ed0068f69fe040ee5ecf308c686f3e96202bd/arches/app/views/api/nodegroup.py#L49-L51

CWDamm-Kint: Done

samuelscandrett (commented):
LGTM.

@CWDamm-Kint does this significantly reduce the time taken for the function to produce GeoJSON from BNG on your local test environment?

@aj-he does this change look okay to you? Just checking there wasn't an intended reason for using `refresh_geojson_geometries` over `refresh_tile_geojson_geometries` in the past?

arches_her/functions/bngpoint_to_geojson_function.py
```diff
@@ -209,10 +209,11 @@ def save_geojson(self, tile, request, is_function_save_method=True):
...
 
             cursor = connection.cursor()
-            sql = """
-                    SELECT * FROM refresh_geojson_geometries();
-                """
-            cursor.execute(sql)  #
+
+            cursor.execute(
+                "SELECT * FROM refresh_tile_geojson_geometries(%s);",
+                (tile.tileid,),
```
aj-he: @samuelscandrett - If you look up, the tile that receives the geometry could be something other that `tile` on ln184. The if/else block starting at ln190 will save it to an existing tile (ln199) or create a new tile ln209). 

The use of `refresh_tile_geojson_geometries` looks good though and should speed things up nicely

arches_her/functions/bngpoint_to_geojson_function.py
samuelscandrett, line 180: We should probably also update the text on this comment from all geojson geoms to just the tile

arches_her/functions/bngpoint_to_geojson_function.py
CWDamm-Kint: Thanks, hopefully this now catches the right tileid, regardless of the logic path it takes. 

I think there might also have been a missing tile.save, for when the bng and the geojson share the same nodegroup, so I've also added that in too.

aj-he (approved):
LGTM

samuelscandrett (approved):
LGTM

